PR #11, opened 2025-03-29
Add ServiceCall factory to account for breaking change in homeasssistant core library

dwats:
### Description
Fixes compatibility issues with ServiceCall instantiations between Home Assistant 2024.x.x and 2025.x.x versions.

In 2025.x.x, the ServiceCall constructor gained a required 'hass' parameter, which broke backward compatibility with integrations directly creating ServiceCall instances. This update adds a factory function that detects the Home Assistant version and creates the appropriate ServiceCall instance format.

### Changes
- Added `create_service_call()` factory function which checks MAJOR_VERSION
- Updated all ServiceCall instantiations to use the factory

### Testing
Tested against Home Assistant 2025.3.x to verify proper ServiceCall creation. Will need a user with 2024.x.x to test to verify as well.

### DIFF
--- a/custom_components/meshcore/services.py
+++ b/custom_components/meshcore/services.py
@@ -3,9 +3,11 @@ import logging
 import time
 import voluptuous as vol
 import json
+from typing import Any, Dict, Optional
 
 from homeassistant.core import HomeAssistant, ServiceCall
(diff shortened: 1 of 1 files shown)


awolden:
Works for me on `2024.x.x`, thanks for fixing this.

PR #13, opened 2025-03-30
Update sensor.py

mikewren:
Changed default unit of measurement for uptime sensor to days rather than minutes.


awolden:
thanks for updating this

PR #35, opened 2025-06-19
2.0.7

awolden:


jquatier:
Would be cool to pull this from the firmware instead now that it’s a board method

PR #77, opened 2025-10-08
Battery Percentage calc

oliverl-21:
Hi,

this adds the same Battery Percentage calculation as the Firmware and APP.


oliverl-21:
yes.

#https://github.com/meshcore-dev/MeshCore/blob/da5dbcd2744d10c15763108276c93652ad98d6d7/examples/companion_radio/ui-new/UITask.cpp#L94

PR #94, opened 2025-11-06
Enhance dashboard overview with logbook card details

WJ4IoT:
In the past I sometimes send a message to the wrong channel, this simple card azures you that the focus of the logbook channel and message channel is the same. But a tiny bit of pre-work is required.


awolden:
Thank you!

Issue #75, opened 2025-10-06
Unable to add monitored repeaters for multiple integratio nentities


dotdavid:
I am unable to add multiple integration entities (meshcore devices, one connected via USB and one via TCP) and have more than one of them able to monitor repeaters.

It works fine if there are two integration entries, and if either one of them is the only one monitoring a repeater station. However if both integration entities are monitoring a repeater station it seems that one of them loses all of its monitored entities in the UI except in the "Managed Monitored Devices" radio button option in the main MeshCore integration page that you get by clicking the gear icon next to each integration entity.

Steps to Reproduce

1. Add two integration entities

<img width="1088" height="551" alt="Image" src="https://github.com/user-attachments/assets/19733752-3e7d-4ceb-b21b-77e8806e7750" />

2. Add a monitored repeater to one of the entities

<img width="1114" height="614" alt="Image" src="https://github.com/user-attachments/assets/e5b535d5-2af3-4531-85df-b3a404ce4a70" />

3. Add a monitored repeater to the other entity. The first monitored repeater disappears from the UI.

<img width="1093" height="565" alt="Image" src="https://github.com/user-attachments/assets/dff2aba7-4f5d-4f8e-bdf1-089e20e28bdf" />

4. The "invisible" monitored repeater still shows in the "Managed Monitored Devices" page

<img width="399" height="406" alt="Image" src="https://github.com/user-attachments/assets/279e4d29-d794-426d-aab8-18f76a0c9396" />

I tried looking at the code but I'm a bit baffled. I assume it's something to do with database keys or something and I do get messages saying "Platform meshcore does not generate unique IDs" in the log for the tracked repeaters. But looking at the code I can't really understand why their IDs wouldn't be unique.

Can anyone reproduce this? There's nothing obvious in the logs that I can see that might be related.

awolden:
There is likely a key collision somewhere. I will have to find another companion node to test it out. Out of curiosity what is the real use case for multiple companion nodes?

PR #131, opened 2026-02-06
Only cleanup monitored repeaters associated with the current integration

mannkind:
Only cleanup the monitored repeaters that belong to the current integration.

I _think_ this should fix https://github.com/meshcore-dev/meshcore-ha/issues/75; it fixed what I was experiencing anyway.

Before the fix:
https://github.com/user-attachments/assets/f70c35d7-f337-4ab2-bfea-9288c07b84ba

After the fix:
https://github.com/user-attachments/assets/70756867-2245-407f-8e72-8117f2df55c0


awolden:
Thanks for fixing this and adding the verification evidence

PR #133, opened 2026-02-07
Update installation.md

frankkienl:
Add note about using the correct companion firmware when using the USB connection. When using the Companion Bluetooth firmware, this it will fail to connect. Tested with Meshcore firmware 1.12.0 on Heltec v3; And `meshcore-ha` 2.2.5

I would very much recommend adding the note to the docs;   
as this may save other people a lot of time on trouble shooting.


awolden (approved):
Useful note. Thanks for catching this.

PR #177, opened 2026-03-28
Subscribe to MESSAGES_WAITING for instant message delivery

mwolter805:
## Summary

- Subscribes to the firmware's `MESSAGES_WAITING` push notification (`0x83`) so incoming messages are fetched immediately instead of waiting for the coordinator's periodic poll
- Adds `async_flush_messages()` to the coordinator with an `asyncio.Lock` to prevent overlap between auto-fetch and the periodic poll's own `get_msg()` loop
- Wraps the existing periodic `get_msg()` block in the same lock for safe serialization

## Problem

When the companion firmware queues incoming messages, it sends a `MESSAGES_WAITING` push notification. The integration never subscribed to this event, so messages sat in the device queue until the coordinator's periodic poll called `get_msg()` — introducing a delay of up to the full poll interval (~3 seconds).

## Changes

| File | Change |
|------|--------|
| `__init__.py` | Register `MESSAGES_WAITING` event subscriber in `async_setup_entry` that calls `coordinator.async_flush_messages()` |
| `coordinator.py` | Add `_message_lock` (asyncio.Lock), new `async_flush_messages()` method, wrap existing periodic message poll in same lock |

## Test plan

- Verified message delivery latency drops from ~3s to near-instant on a live HA host with MeshCore companion device
- Confirmed periodic poll still works correctly alongside auto-fetch
- No regressions in existing message handling

### DIFF
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -492,6 +492,21 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             handle_new_contact
         )
 
+        # Subscribe to MESSAGES_WAITING for instant message delivery.
+        # The companion firmware sends this push notification when messages
+        # are queued on the device.  Without this, messages sit in the device
+        # queue until the coordinator's periodic poll calls get_msg().
+        async def handle_messages_waiting(event):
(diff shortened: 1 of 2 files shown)


awolden (approved):
looks good but has merge conflicts

PR #182, opened 2026-03-30
Replace blocking RX_LOG correlation sleep with adaptive poll-wait

mwolter805:
### Summary

Incoming channel messages currently block for a fixed 500ms (`asyncio.sleep(0.5)`) while waiting for RX_LOG correlation data before firing the `meshcore_message` event. This delays every channel message by half a second regardless of whether RX_LOG data arrives sooner.

This PR adds an **opt-in** adaptive poll-wait mode, controlled by a new config toggle in Global Settings. **The default behavior is unchanged** — the fixed 500ms sleep is preserved out of the box.

### Config Option

**Global Settings → Adaptive Channel Message Delivery** (disabled by default)

- **Disabled (default):** Fixed 500ms sleep, then attach whatever RX_LOG data accumulated. Identical to current `main` behavior. No changes needed for existing consumers.
- **Enabled:** Poll `_pending_rx_logs` every 50ms and fire as soon as data arrives (500ms ceiling). A background task collects late-arriving repeater RX_LOGs and delivers them via `meshcore_delivery_update` events.

### Changes

- `const.py` — New `CONF_ADAPTIVE_POLL_WAIT` constant
- `config_flow.py` — Added toggle to `async_step_global_settings()` (save, read, form schema)
- `logbook.py` — `handle_channel_message()`: config-driven branch between fixed-wait (default) and adaptive poll-wait. New constants: `_INCOMING_FIXED_WAIT` (0.5s), `_INCOMING_POLL_INTERVAL` (0.05s), `_INCOMING_MAX_POLLS` (10)
- `translations/en.json` — Label and description for the new option

### Behavior

- **Default (adaptive disabled):** Identical to current `main`. Every incoming channel message waits 500ms. `rx_log_data` attached to initial event.
- **Adaptive enabled:** `meshcore_message` fires as soon as RX_LOG data arrives (typically ~50ms), with a 500ms ceiling. Late-arriving repeater data delivered progressively via `meshcore_delivery_update` events.

**No breaking changes.** Default behavior is unchanged. The adaptive mode is strictly opt-in.

### Testing

Tested on a live HA host with a MeshCore network. Results across 3 channel messages routed through multiple repeaters (adaptive mode enabled):

| Data Arrived | Route | Latency Reduction |
|-------------|-------|-------------------|
| 50ms (poll 1/10) | 5 repeaters | 500ms → 50ms |
| 50ms (poll 1/10) | 4 repeaters | 500ms → 50ms |
| 50ms (poll 1/10) | 3 repeaters | 500ms → 50ms |

RX_LOG data arrived on the first poll in all cases, even for messages routed through up to 5 repeaters. Background collection passes found no additional late-arriving data. No errors in HA logs.

### DIFF
--- a/custom_components/meshcore/config_flow.py
+++ b/custom_components/meshcore/config_flow.py
@@ -51,6 +51,7 @@ from .const import (
     CONF_SELF_TELEMETRY_INTERVAL,
     DEFAULT_SELF_TELEMETRY_INTERVAL,
     CONF_MAP_UPLOAD_ENABLED,
+    CONF_ADAPTIVE_POLL_WAIT,
     CONF_MQTT_IATA,
     CONF_MQTT_TOKEN_TTL_SECONDS,
     CONF_MQTT_BROKERS,
@@ -880,6 +881,7 @@ class OptionsFlowHandler(config_entries.OptionsFlow):
             new_data[CONF_SELF_TELEMETRY_ENABLED] = user_input[CONF_SELF_TELEMETRY_ENABLED]
             new_data[CONF_SELF_TELEMETRY_INTERVAL] = user_input[CONF_SELF_TELEMETRY_INTERVAL]
             new_data[CONF_MAP_UPLOAD_ENABLED] = user_input[CONF_MAP_UPLOAD_ENABLED]
+            new_data[CONF_ADAPTIVE_POLL_WAIT] = user_input[CONF_ADAPTIVE_POLL_WAIT]
             self.hass.config_entries.async_update_entry(self.config_entry, data=new_data) # type: ignore
 
             if new_data[CONF_LIMIT_DISCOVERED_CONTACTS]:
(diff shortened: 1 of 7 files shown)


mwolter805:
Thanks for the feedback. I've reworked the approach so it's not a breaking change.

Instead of removing the sleep entirely, the initial `meshcore_message` event now polls for RX_LOG data every 50ms, up to the same 500ms ceiling as before. As soon as data arrives, it fires immediately — so `rx_log_data` is still present on the initial event, same as today. In testing, data consistently arrived on the first poll (~50ms), so typical latency drops from 500ms to ~50ms while worst-case behavior is unchanged.

On top of that, a background task now collects late-arriving repeater RX_LOGs (two passes at 0.5s and 1.0s after the initial fire) and delivers them via `meshcore_delivery_update` events — the same progressive pattern already used for outgoing messages. This is purely additive; existing consumers don't need any changes.

PR #201, opened 2026-04-07
Fix entity identity model violation and config change propagation

mwolter805:
## Summary

Six entity classes include the device name in their `unique_id`, which violates Home Assistant's entity identity model. HA requires `unique_id` to be immutable for the lifetime of an entity — it's the permanent key in the entity registry. Because device names can change (via `set_name`, repeater renames, etc.), any name change either orphans existing entities or creates duplicates, depending on how reconstruction happens.

This PR stabilizes `unique_id` by removing the name component, ensures runtime config changes propagate to sensors immediately (no restart required), and fixes missing state-write calls in six SELF_INFO sensor handlers.

A detailed forensics report documenting the original investigation is available here: https://gist.github.com/mwolter805/060566596c3951d45f5acdcf90293dc6

### Entity identity violation

HA's [entity registry documentation](https://developers.home-assistant.io/docs/entity_registry_index/#unique-id-requirements) states that `unique_id` must be stable and never change. The following entity classes currently include the device name in their `unique_id`:

| Entity Class | Current `unique_id` | Fixed `unique_id` |
|---|---|---|
| `MeshCoreSensor` | `{entry_id}_{key}_{pubkey[:6]}_{name}` | `{entry_id}_{key}_{pubkey[:6]}` |
| `RateLimiterSensor` | `{entry_id}_rate_limiter_tokens_{pubkey[:6]}_{name}` | `{entry_id}_rate_limiter_tokens_{pubkey[:6]}` |
| `LastMessageDeliverySensor` | `{entry_id}_last_message_delivery_{pubkey[:6]}_{name}` | `{entry_id}_last_message_delivery_{pubkey[:6]}` |
| `MeshCoreReliabilitySensor` | `{entry_id}_{type}_{pubkey}_{key}_{pubkey[:6]}_{name}` | `{entry_id}_{type}_{pubkey}_{key}_{pubkey[:6]}` |
| `MeshCorePathSensor` | `{entry_id}_{type}_{pubkey}_{key}_{pubkey[:6]}_{name}` | `{entry_id}_{type}_{pubkey}_{key}_{pubkey[:6]}` |
| `MeshCoreRepeaterSensor` | `{entry_id}_repeater_{pubkey}_{key}_{pubkey[:6]}_{name}` | `{entry_id}_repeater_{pubkey}_{key}_{pubkey[:6]}` |

When a user renames their companion device or a tracked repeater changes its advertised name, HA sees a "new" entity (different `unique_id`) and orphans the old one. Over time this silently accumulates dead entities in the registry.

### No SELF_INFO refresh after config commands

When a user changes device configuration at runtime via `execute_command` (e.g., `set_radio`, `set_tx_power`, `set_name`, `set_coords`), the firmware applies the change immediately. But the integration never requests the device's updated configuration afterward, so HA sensors continue showing stale values until the next restart or reload.

### Missing `async_write_ha_state()` in sensor handlers

Six SELF_INFO-driven sensors (`tx_power`, `latitude`, `longitude`, `frequency`, `bandwidth`, `spreading_factor`) update their internal `_native_value` when a SELF_INFO event arrives, but never call `async_write_ha_state()` to notify HA. The updated values sit in memory until an unrelated coordinator cycle happens to trigger a state write. Dashboard cards and automations see stale data for an unpredictable duration.

### 1. `unique_id` stabilization (one-time migration)

`_migrate_unique_ids_remove_name()` runs at startup and strips the device name suffix from all entity `unique_id` values in the entity registry. After migration, `unique_id` uses only stable identifiers (`entry_id`, `key`, `pubkey`) and survives name changes without duplication or orphaning.

The migration is safe to re-run (idempotent) and checks for collisions before updating.

### 2. SELF_INFO refresh after config commands

A `_SELF_INFO_COMMANDS` set identifies commands that modify values reported in SELF_INFO (`set_radio`, `set_tx_power`, `set_name`, `set_coords`, and others). After any of these succeed, `send_appstart()` is called to refresh the cached SELF_INFO so sensors pick up the new values immediately.

### 3. `async_write_ha_state()` added to 6 sensor handlers

The `tx_power`, `latitude`, `longitude`, `frequency`, `bandwidth`, and `spreading_factor` SELF_INFO event handlers now call `self.async_write_ha_state()` after updating `_native_value`, matching the pattern used by other sensors in the same codebase (`node_count`, `CompanionPrefixSensor`).

## Changes

| File | What changed |
|---|---|
| `__init__.py` | Added `_migrate_unique_ids_remove_name()` idempotent startup migration |
| `sensor.py` | Removed name from `unique_id` in 6 entity classes, added `async_write_ha_state()` to 6 SELF_INFO handlers |
| `services.py` | Added `_SELF_INFO_COMMANDS` set and post-command `send_appstart()` refresh |

Net diff against `main`: 3 files, +137 / -15.

## Testing

Tested on a live HA host with a MeshCore companion device and multiple tracked repeaters:

- Changed radio frequency via `set_radio` — frequency sensor updated within seconds (previously required restart)
- Exercised each of the 6 SELF_INFO sensors — all wrote state immediately after the value changed
- Verified one-time `unique_id` migration stripped name suffix from all affected entities on first startup after update
- Confirmed migration is idempotent — no changes on subsequent restarts
- Renamed companion via `set_name` — display name updates via `has_entity_name`; entity_ids stay put (intentional); no duplication or orphaning, since `unique_id` no longer carries the name

## Known limitation: pubkey in unique_id

The companion device's `pubkey[:6]` (and in some entity classes the full `pubkey`) remains in `unique_id` after this PR. This is also technically a mutable value — a firmware reflash or factory reset changes the public key. However, this is already mitigated by the existing `_migrate_entity_ids()` function, which detects pubkey changes at startup and rewrites all entity IDs and unique IDs to use the new prefix.

For companion entities, `entry_id` alone already uniquely scopes to one device, making the pubkey in `unique_id` redundant. Removing it would make companion entity `unique_id` values truly immutable. For tracked repeaters/clients, pubkey is structurally necessary to distinguish entities under the same config entry. This could be addressed in a follow-up PR — scoped out of this one to keep the change focused on the name-related identity violation and config propagation fixes.

## No breaking changes

- The `unique_id` migration is transparent — HA preserves entity history, customizations, and automation references through `unique_id` updates
- Entity IDs (the user-facing identifiers) are unchanged by this PR. `has_entity_name` continues to update the display name when a device is renamed; the entity_id itself stays put. This is intentional — silently renaming entity_ids would break existing automations, dashboards, and scripts.

### DIFF
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -26,6 +26,7 @@ from homeassistant.helpers import issue_registry as ir
 from .const import (
     DOMAIN,
     CONF_CONNECTION_TYPE,
+    CONF_NAME,
     CONF_USB_PATH,
     CONF_BLE_ADDRESS,
     CONF_TCP_HOST,
@@ -38,6 +39,7 @@ from .const import (
     DEFAULT_MAX_DISCOVERED_CONTACTS,
     CONF_MESSAGES_INTERVAL,
     DEFAULT_UPDATE_TICK,
+    REPAIR_PUBKEY_CHANGED,
 )
 from .coordinator import MeshCoreDataUpdateCoordinator
 from .meshcore_api import MeshCoreAPI
@@ -161,6 +163,90 @@ def _migrate_entity_ids(
     )
 
 
+def _migrate_unique_ids_remove_name(
+    hass: HomeAssistant,
+    entry: ConfigEntry,
+) -> None:
+    """One-time migration: remove device name from unique_ids.
+
+    Entity unique_ids previously included the device name, which made them
+    unstable across name changes. This migration strips the name suffix
+    so unique_ids use only stable identifiers (entry_id, key, pubkey).
+    """
+    entity_registry = er.async_get(hass)
+    migrated = 0
+
+    # Get the current companion name from config entry
+    companion_name = entry.data.get(CONF_NAME, "")
+
+    # Build list of all names that might appear in unique_ids
+    # (companion + all tracked repeaters + all tracked clients).
+    # Sorted longest-first so that if one name is a suffix of another
+    # (e.g., "mytest" vs "test"), the longer name matches first via
+    # endswith(), preventing partial stripping.
+    names_raw: set[str] = set()
+    if companion_name:
+        names_raw.add(companion_name)
+    for sub in entry.data.get(CONF_REPEATER_SUBSCRIPTIONS, []):
+        name = sub.get("name", "")
+        if name:
+            names_raw.add(name)
+    for sub in entry.data.get(CONF_TRACKED_CLIENTS, []):
+        name = sub.get("name", "")
+        if name:
+            names_raw.add(name)
+
+    names_to_strip = sorted(names_raw, key=len, reverse=True)
+
+    if not names_to_strip:
+        return
+
+    for entity in list(entity_registry.entities.values()):
+        if entity.config_entry_id != entry.entry_id:
+            continue
+
+        new_unique_id = entity.unique_id
+        for name in names_to_strip:
+            # unique_ids have the name appended with underscore separator
+            suffix = f"_{name}"
+            if new_unique_id.endswith(suffix):
+                new_unique_id = new_unique_id[: -len(suffix)]
+                break
+
+        if new_unique_id == entity.unique_id:
+            continue
+
+        # Verify the new unique_id doesn't already exist
+        existing = entity_registry.async_get_entity_id(
+            entity.domain, DOMAIN, new_unique_id
+        )
+        if existing and existing != entity.entity_id:
+            _LOGGER.warning(
+                "Cannot migrate unique_id for %s: target %s already exists (%s)",
+                entity.entity_id, new_unique_id, existing,
+            )
+            continue
+
+        try:
+            entity_registry.async_update_entity(
+                entity.entity_id,
+                new_unique_id=new_unique_id,
+            )
+            _LOGGER.info(
+                "Stabilized unique_id: %s -> %s",
+                entity.unique_id, new_unique_id,
+            )
+            migrated += 1
+        except Exception as ex:
+            _LOGGER.error(
+                "Failed to stabilize unique_id for %s: %s",
+                entity.entity_id, ex,
+            )
+
+    if migrated:
+        _LOGGER.info("Stabilized %d entity unique_ids (removed name)", migrated)
+
+
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     """Set up MeshCore from a config entry."""
     # Home Assistant can trigger a duplicate setup during rapid reload/update cycles.
@@ -254,7 +340,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             is_fixable=False,
             is_persistent=True,
             severity=ir.IssueSeverity.WARNING,
-            translation_key="pubkey_changed",
+            translation_key=REPAIR_PUBKEY_CHANGED,
             translation_placeholders={
                 "old_key": stored_pubkey[:12],
                 "new_key": live_pubkey[:12],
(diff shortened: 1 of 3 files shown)


awolden:
The unique_id fix is the right call. Names in unique_ids cause orphaned entities on rename and that needs to be fixed. The one-time migration to strip names from unique_ids is good, and since it's idempotent it's safe to run on every startup without needing version gating.

I think we should drop the entity_id rename machinery though (the runtime name change detection, `_async_migrate_entity_name`, repair issues, SELF_INFO refresh, etc). That's about 75% of this PR and adds a lot of moving parts to core components.

Names in entity_ids are intentional and important for usability. `sensor.meshcore_hilltop_repeater_battery` is immediately recognizable vs `sensor.meshcore_abc123_battery`. HA encourages human-readable entity_ids through the `has_entity_name` pattern, and we should keep that. But entity_ids should be treated as immutable by the integration since renaming them silently breaks automations, dashboards, and scripts. If a user renames their device, the display name updates via `has_entity_name`, and the entity_id keeps the original name. If they want fresh entity_ids they can remove and re-add the device.

Can we scope this down to just:
1. Strip names from unique_ids in sensor.py
2. The idempotent startup migration to fix existing unique_ids
3. The missing `async_write_ha_state()` calls (independent bug fix)

That gets us the stability fix without the risk of the runtime rename logic.

mwolter805:
Agree on scoping down the entity_id rename machinery — that adds complexity to core components and the has_entity_name pattern handles display name updates without touching entity_ids. I'll pull that out along with _async_migrate_entity_name(), the name change detection, and the repair issues.

I'd like to make the case for keeping the SELF_INFO refresh (send_appstart() after config commands) though. Right now when a user runs set_radio, set_tx_power, set_coords, etc., the firmware applies the change immediately but the integration never fetches the updated state. Sensors continue showing the old values until the next restart or reload. This is confusing because other MeshCore interfaces (the companion app, the serial console) reflects these changes instantly. Users see the command succeed but HA still shows stale data — it looks like the command didn't work.

The SELF_INFO refresh and the async_write_ha_state() fixes are really two halves of the same config propagation bug: SELF_INFO refresh gets the new values into memory, and async_write_ha_state() pushes them to HA. Without both, sensors still show stale data after config commands.

So the scoped PR would be:

- Strip names from unique_id in sensor.py
- Idempotent startup migration for existing unique_id values
- async_write_ha_state() calls in the 6 SELF_INFO sensor handlers
- _SELF_INFO_COMMANDS + send_appstart() refresh after config commands

Would that work?

awolden:
Good point on SELF_INFO, should stay in, it's valuable.

mwolter805:
Ok great, then the last force-push should be good. Let me know if it needs anything else.

awolden:
Thanks for the updates. The unique_id strip, migration, async_write_ha_state fixes, and SELF_INFO refresh all look good.

I think the last force-push might have squashed rather than rescoped. The rename machinery is still in the diff:

- `_async_migrate_entity_name` and the SELF_INFO name-change subscriber in `__init__.py`
- `_async_check_tracked_node_name_changes` and `on_tracked_node_name_change` in `coordinator.py`
- `REPAIR_NAME_CHANGED` / `REPAIR_TRACKED_NODE_NAME_CHANGED` in `const.py` plus the translation strings

The commit message for `6fe089b` also still lists (C) "Detect name changes, migrate entity IDs, create repair issues" which matches what's in the code.

Can you drop those and keep only (A) SELF_INFO refresh, (B) async_write_ha_state, and (D) unique_id strip + migration? Once that's out it's ready to ship.

mwolter805:
You were right — that was a squash, not a rescope. I've force-pushed again with the rename machinery actually removed.
What's in now:

- SELF_INFO refresh after config commands (_SELF_INFO_COMMANDS + send_appstart() in services.py)
- async_write_ha_state() added to the 6 SELF_INFO sensor handlers in sensor.py
- unique_id strip + one-time idempotent migration (sensor.py + __init__.py)

What's out:

- _async_migrate_entity_name and the SELF_INFO name-change subscriber — gone from __init__.py
- _async_check_tracked_node_name_changes and on_tracked_node_name_change — gone from coordinator.py
- REPAIR_NAME_CHANGED / REPAIR_TRACKED_NODE_NAME_CHANGED — gone from const.py along with the matching translation strings

Diff is now 3 files (__init__.py, sensor.py, services.py), +137 / -15 against main. const.py, coordinator.py, and translations/en.json are untouched. Commit message updated to reflect the narrower scope. Let me know if anything else stands out.

awolden (approved):
Scope matches now. Thanks for the re-push. Shipping.

PR #209, opened 2026-04-19
Add per-device online binary sensor for managed repeaters and clients

mwolter805:
### Summary

Adds a per-device connectivity binary sensor (`binary_sensor.meshcore_<pubkey>_online_<name>`) for every managed repeater and client. It reports whether that *specific remote node* is currently being polled successfully — which is different from whether the companion device's USB/TCP/BLE link is up.

### Why

The existing `sensor.meshcore_..._node_status` entity reports the **companion** device's connection state — i.e., is the LoRa radio itself reachable from Home Assistant. That's useful, but it's not a per-node health signal: a tracked repeater or client can stop responding (out of range, powered off, degraded routing path) while the companion stays connected, and today there is no cheap way to alert on that.

This PR adds a dedicated sensor per managed device with a staleness-based definition of "online," so users can build the obvious automations (notify when node X goes silent, dashboard cards that grey-out offline nodes, conditional routing in scripts) without custom templates.

The sensor is consumed by the upcoming MeshCore sidebar panel for its connection-state indicators, but it's a fully native HA entity on its own — no frontend required to use it.

### What changes

**New entity (one per tracked repeater/client):**

- `binary_sensor.meshcore_<pubkey>_online_<name>`
- Device class: `connectivity`
- States:
  - **On** — last successful poll for this device was within the staleness window
  - **Off** — companion device is disconnected, OR no successful poll within the staleness window
  - **Unknown** — device has never been successfully polled this session

**Staleness window:** `2.5 × max(device_update_interval, 300s)` — scales with each device's configured refresh rate. A node on the default 5-minute update cadence becomes "off" after 12.5 minutes of no successful polls. A device configured to poll every 10 minutes gets a 25-minute window.

**Attributes:**

- `last_successful_request` — ISO 8601 timestamp of the most recent successful poll
- `update_interval` — device's configured telemetry refresh rate (seconds)
- `staleness_window` — current threshold (seconds)

This series of PRs (along with `feature/message-store` and `feature/repeater-neighbors`) exists mainly to support the upcoming sidebar-panel frontend work. That said, each new entity, service, or attribute is a first-class HA object and is fully usable from YAML automations, templates, scripts, and REST/WebSocket consumers. You do not need the sidebar panel to benefit from this PR.

Concrete non-frontend use cases for this sensor:

- **Alerting** — trigger a notification when a critical repeater goes `off` for more than N minutes.
- **Conditional automations** — only run an automation if the target node has been `on` recently.
- **Dashboard health grids** — colour entity cards by `state` of the online sensor across all nodes.
- **Template fan-in** — count online vs total managed devices and publish a mesh-uptime percentage sensor.

### Documentation

Adds a "Device Online Status" section to `docs/docs/sensors.md` under Binary Sensors, covering the entity naming pattern, state semantics, staleness formula, and the distinction vs `sensor.meshcore_..._node_status`.

### Related / upcoming

Companion PRs planned from this work: `feature/message-store` (persistent per-conversation message store), `feature/repeater-neighbors` (repeater neighbor SNR/activity sensors). The sidebar-panel frontend that consumes all three is tracked separately and will follow.

### DIFF
--- a/custom_components/meshcore/binary_sensor.py
+++ b/custom_components/meshcore/binary_sensor.py
@@ -26,6 +26,7 @@ from .const import (
     MESSAGES_SUFFIX,
     CHANNEL_PREFIX,
     CONTACT_SUFFIX,
+    ONLINE_SUFFIX,
     NodeType,
 )
 from .utils import (
@@ -259,6 +260,21 @@ async def async_setup_entry(
         ]
         if mqtt_entities:
             async_add_entities(mqtt_entities)
+
+    # Create online status binary sensors for managed devices
+    online_entities = []
+    for repeater_config in coordinator._tracked_repeaters:
+        if repeater_config.get("pubkey_prefix"):
+            online_entities.append(
+                MeshCoreDeviceOnlineBinarySensor(coordinator, repeater_config, "repeater")
+            )
+    for client_config in coordinator._tracked_clients:
+        if client_config.get("pubkey_prefix"):
+            online_entities.append(
+                MeshCoreDeviceOnlineBinarySensor(coordinator, client_config, "client")
+            )
+    if online_entities:
+        async_add_entities(online_entities)
 
     # Subscribe to our internal message sent event for outgoing messages
     @callback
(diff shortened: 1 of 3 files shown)


awolden (approved):
Clean implementation. State transitions (Unknown → Off → On) look correct, staleness window correctly uses per-device update interval, no conflation of available/is_on. Minor nits (reuse `build_device_id` from utils.py, small wording tweak to `update_interval` doc) are non-blocking. Shipping.

PR #210, opened 2026-04-19
Add per-conversation message store with search and bounded retention

mwolter805:
### Summary

Adds a purpose-built, per-conversation message store — independent of the Home Assistant logbook — that records each message's text, sender, timestamp, delivery status, and RX_LOG routing data. Exposes two new services (`meshcore.get_messages`, `meshcore.search_messages`) and an internal WebSocket API used by the upcoming sidebar panel.

### Why the logbook isn't sufficient

The logbook is fine as a surface for individual message *events*, but it doesn't work as a backing store for a conversation UI, and it isn't ergonomic for automations that need structured per-message context:

- **No unread tracking.** The logbook has no concept of read state. A conversation UI must know which messages the user has and has not seen since the last visit — that has to live somewhere persistent.
- **RX_LOG correlation is lost.** When a channel message is received, the integration correlates multiple RX_LOG entries (each with SNR, RSSI, hop count, routing path) onto the message event as `rx_log_data`. The logbook only captures the summary; the routing detail scrolls off the event bus. The message store keeps `rx_log_data` attached to each message.
- **Delivery status is asynchronous.** For outgoing messages, delivery acks and round-trip timing arrive after the initial send. The store updates the existing record in place — the logbook would need a second entry.
- **Bounded retention, predictable footprint.** The logbook grows with overall HA activity and is tuned for the whole system. Conversation history should roll off on its own schedule, independent of the recorder's settings.
- **Efficient scrollback.** Loading a conversation in the sidebar is a single JSON read, not a recorder query across months of logbook rows.

### Size and footprint — the 500-message cap is load-bearing

The store is deliberately small. The limits are:

| Setting | Default | Notes |
|---|---|---|
| Messages per conversation | **500** | FIFO drop on overflow. Configurable via options. |
| Retention window | **90 days** | Daily cleanup; configurable. |
| Idle eviction | **5 minutes** | Conversations not accessed for 5 minutes are flushed to disk and dropped from memory. |

The 500-message ceiling matters: a busy public channel hits that cap well before it hits 90 days, so each conversation file naturally stays in single-digit megabytes even on active networks. Unused conversations cost zero memory — they live on disk until opened.

Each conversation is one JSON file under `.storage/` (`meshcore.<entry_id>.msgs.<entity_id>`). A lightweight index (`meshcore.<entry_id>.msgidx`) holds one row per conversation with count, last sender, last timestamp, and a 50-character preview — enough for the sidebar's conversation list to render without loading any message bodies.

Like the other PRs in this series, the proximate reason for this work is the upcoming sidebar-panel frontend ([discussion on Discord](https://discord.com/channels/1343693475589263471/1358879320248160419/1489089455045476483)) — specifically its conversation list, scrollback, and unread-indicator behaviour. But the two services and the per-record structure are first-class HA surfaces, useful on their own:

- **Activity scans.** `search_messages` lets a template or automation flag whether a keyword has appeared in any conversation in recent history.
- **Delivery verification.** After sending a message, call `get_messages` against the destination entity and inspect `delivery_status` on the latest record.
- **Mesh-path analysis.** `rx_log_data` on each incoming message lists every repeater that relayed it, with per-hop SNR and RSSI — usable directly in a template sensor or `python_script` to track routing diversity over time.
- **Bot-style integrations.** A bot can tail a conversation entity, match on message text, and reply by calling `meshcore.send_message` — no need to stand up a separate transcript database.

### What changes

**Services:**

- `meshcore.get_messages` — retrieve stored messages for a conversation. Fields: `entity_id` (required), `limit` (1–500, default 50), `before` (pagination cursor), optional `entry_id` for multi-device setups. Returns a structured response.
- `meshcore.search_messages` — substring search across text and sender name. Fields: `query` (required, case-insensitive), optional `entity_id` to limit scope, `limit` (1–100, default 20). Returns a structured response.

**Storage layer:**

- Per-conversation files plus an integration-level index, stored via HA's `Store` helper (backup-compatible, restart-durable).
- Fire-then-enhance: outgoing messages are written locally before the wire send completes; `delivery_status` updates in place when the ack lands; `rx_log_data` is attached progressively as repeater relays report in.
- Lazy conversations — no "enable" switch, no upfront allocation; a conversation file is created the first time a message is stored for that entity.

### Documentation

New page: `docs/docs/message-store.md` (sidebar position 5.5), covering the "why" motivation, retention/eviction semantics, the two services with schemas and example YAML, and automation use cases.

### Related / upcoming

Companion PRs from this series: `feature/per-device-online-sensor` (per-device connectivity sensor), `feature/repeater-neighbors` (repeater neighbor SNR/activity sensors). The sidebar-panel frontend that consumes all three is tracked separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ script_venv
 __pycache__/
 *.pyc
 .pytest_cache/
-node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ script_venv
 __pycache__/
 *.pyc
 .pytest_cache/
+node_modules/
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -44,6 +44,7 @@ from .meshcore_api import MeshCoreAPI
 from .map_uploader import MeshCoreMapUploader
 from .mqtt_uploader import MeshCoreMqttUploader
 from .services import async_setup_services, async_unload_services
+from .ws_api import async_register_ws_commands
 from .utils import (
     create_message_correlation_key,
     parse_and_decrypt_rx_log,
@@ -362,7 +363,14 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     
     # Set up all platforms for this device
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
-    
+
+    # Initialize persistent message store
+    await coordinator.message_store_startup()
+    await coordinator.cleanup_old_messages()
+
+    # Register WebSocket commands for message store
+    async_register_ws_commands(hass)
+
     # Register static paths for icons
     should_cache = False
     icons_path = Path(__file__).parent / "www" / "icons"
@@ -561,6 +569,9 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             for remove_listener in coordinator._remove_listeners:
                 remove_listener()
                 
+        # Flush message store before disconnect
+        await coordinator.flush_message_stores()
+
         # Disconnect from the device
         if getattr(coordinator, "mqtt_uploader", None):
             await coordinator.mqtt_uploader.async_stop()
--- a/custom_components/meshcore/const.py
+++ b/custom_components/meshcore/const.py
@@ -146,6 +146,12 @@ CONF_ADAPTIVE_POLL_WAIT: Final = "adaptive_poll_wait"
 # Sensor availability timeout multiplier
 SENSOR_AVAILABILITY_TIMEOUT_MULTIPLIER: Final = 3
 
+# Message store settings
+DEFAULT_MESSAGE_RETENTION_DAYS: Final = 90
+DEFAULT_MAX_MESSAGES_PER_CONVERSATION: Final = 500
+MESSAGE_STORE_SAVE_DELAY_SECONDS: Final = 5.0
+MESSAGE_STORE_IDLE_EVICTION_SECONDS: Final = 300  # 5 minutes
+
 
 class NodeType(IntEnum):
     CLIENT = 1
--- a/custom_components/meshcore/coordinator.py
+++ b/custom_components/meshcore/coordinator.py
@@ -5,7 +5,7 @@ import asyncio
 import logging
 import random
 import time
-from datetime import timedelta
+from datetime import datetime, timedelta
 from typing import Any, Dict
 
 from cachetools import TTLCache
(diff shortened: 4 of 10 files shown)


awolden:
Stepping back on this one before going through the code in detail.

This PR is a noticeable shift in what the integration owns. Most of meshcore-ha sticks to device I/O (reading from the companion, surfacing entities, firing events). The one existing exception is the discovered-contacts system, which is fair, but that's lightweight compared to a full conversation store with retention, eviction, search, and a WS API. This starts to look more like an application layer than an integration.

A few questions before I dig into the implementation:

1. **Where does the panel live?** You mentioned back in April that it was a "bolt on front end" to meshcore-ha. Is that still the plan, or has the thinking evolved toward shipping it as a companion integration (like Ratty's MeshCore-HA-UI is doing)? That changes the answer here a lot.

2. **How critical is the store living in meshcore-ha specifically?** Ratty's integration handles its own message persistence independently, talking to meshcore-ha over the existing WS API and event stream. If your panel did the same, the persistence layer could live with the panel code and meshcore-ha stays focused on the device. Is there a capability the companion-integration model can't cover?

3. **Have you looked at alternatives?** Recorder plus a small read-marker state covers scrollback and unread for most cases without a new store. Not as clean as what you've built, but much smaller surface area. Have you ruled that out, and if so what drove it?

Not trying to kill this. You've clearly put real work in, and the code itself is well-structured. Just want to be sure we're making the right call on where the responsibility lives before committing to maintain it in meshcore-ha long term.

mwolter805:
Thanks for the pushback — you're right that this crosses from device I/O into application territory, and the maintenance argument lands.

After thinking it through, I'm going to pull the message store and the panel work out of meshcore-ha and ship them as a separate companion integration — MeshCore Home Assistant Chat (meshcore_chat domain, installed via HACS alongside meshcore-ha). That keeps meshcore-ha focused on the radio and the entity/event surface, and the persistence + UI concerns live somewhere you don't have to own. I'll close #210 once the companion repo is up so nothing dangles here.

Direct answers to your three questions:

- Where the panel lives — as a companion, per Ratty's model. That was the right call.
- How critical is the store living in meshcore-ha — it isn't. The companion can own persistence by subscribing to meshcore_message / meshcore_delivery_update and calling the existing send services. The extras I wanted (per-conversation files, idx for incremental load, retention, search) all work fine on the companion side.
- Recorder + read-marker — valid for scrollback and unread alone. It doesn't cover correlating route/SNR/RSSI metadata with each message, which is the part I cared about. Small-surface-area argument still holds; though I'm not going to pretend the companion store is lighter.

I'll be sending a few small PRs your way to smooth the communication between meshcore-ha and the companion. None of them pull app-layer code back in — they're all about making the public interface a bit more complete for any companion:

- Add hop_count and snr to meshcore_message for incoming DMs (they're already on channel events; the DM path drops them).
- A short docs/companion-integration-api.md documenting the events, services, and entity-ID conventions a companion can depend on — purely descriptive of what already exists.
- Structured meshcore.get_contacts / get_channels / trace services as an alternative to scraping execute_command output. Happy to drop this one if you'd rather not expand the service surface.

Each will come as its own focused PR with a clear justification. If any don't fit where you want the public interface to go, easy to abandon individually.
If the companion proves out and you ever want to revisit bringing it back in, happy to have that conversation down the road. For now, best outcome: meshcore-ha stays lean, chat concerns live outside your maintenance burden, and the interface between the two is small and documented.

Appreciate the thoughtful review — much better result than if this had landed as-is.

awolden:
Good pivot. All three follow-ups are fine with me. Looking forward to the companion repo.

PR #211, opened 2026-04-19
Add repeater neighbor SNR and activity sensors with auto-cleanup

mwolter805:
### Summary

Each repeater in a MeshCore network keeps an in-firmware list of the other nodes it has recently heard. This PR surfaces that list to Home Assistant as a pair of sensors per neighbor — one for signal quality, one for activity — giving a first-order view of mesh topology and health from the perspective of each repeater.

The feature is **disabled by default** on every repeater and must be turned on explicitly per-repeater.

### Why

Today the integration has no view into *who each repeater is hearing*. You can infer some of this by correlating RX_LOG entries yourself, but that's heavy-handed for anything an operator would want from a dashboard. Exposing the firmware's own neighbor list directly gives a bounded, structured view of local radio environment per repeater — which is exactly the information needed to answer "is this repeater still hearing the nodes it used to hear" or "did we pick up a new node nearby."

The sensors are consumed by the upcoming sidebar-panel frontend for its mesh-health views, but they are fully native HA entities usable on their own for automations, alerts, and long-term graphing.

### What changes

**Two sensors per neighbor, attached to that repeater's existing HA device:**

- **SNR sensor** — `sensor.meshcore_<repeater>_neighbor_<neighbor>`
  - Unit: dB, state class: measurement
  - Value: most recent signal-to-noise ratio the repeater measured from that neighbor
  - Availability: unavailable if the neighbor has not been heard in the last 72 hours

- **Activity sensor (`seen_48h`)** — `sensor.meshcore_<repeater>_neighbor_<neighbor>_seen`
  - Unit: count, state class: measurement (the value decreases as old sightings age out of the 48h window)
  - Value: number of times this neighbor was heard by the repeater in the last 48 hours
  - Availability: same 72-hour threshold

Friendly names resolve to the neighbor's advertised name when available, with a public-key prefix fallback. Names update if the neighbor later appears in the contact list.

**Per-repeater enable toggle** (disabled by default):

Settings → Devices & Services → Meshcore → Configure → open the per-repeater subscription → toggle **Enable Neighbor Entities**. On the next repeater update cycle the integration queries firmware for the neighbor list and creates sensors. Disabling the toggle removes all neighbor sensors for that repeater, clears in-memory state, and deletes persisted data.

**Persistence and restart-robustness:**

Neighbor data persists via HA's `Store` helper. After a restart, sensors are recreated from disk and the 48-hour activity window continues seamlessly. The integration guards `seen_timestamps` against inflated `secs_ago` values that firmware reports during its own startup, so the rolling window stays accurate across both HA and repeater reboots.

**Integration-level auto-cleanup (also disabled by default):**

A daily task can remove neighbors that haven't been heard within a configurable threshold. Configurable under Configure → Global Settings:

- **Auto-Remove Stale Neighbors** — on/off (default off)
- **Stale Neighbor Threshold (days)** — 1 to 365, default 7

Prevents unbounded neighbor growth in mobile or dense deployments where nodes come and go.

As with the other PRs in this series, the direct motivation is the upcoming sidebar-panel frontend. But the SNR and `seen_48h` sensors are first-class HA entities and are useful outside of any frontend:

- **Mesh health at a glance.** Each repeater's neighbor SNR sensors form a snapshot of its local radio environment. Graphing over time shows degradation, new arrivals, or links dropping out before they become network-wide problems.
- **Activity-based alerts.** `seen_48h` sensors make it easy to flag a repeater that has stopped hearing a neighbor it normally hears dozens of times a day.
- **Coverage mapping.** Combining neighbor lists from multiple repeaters gives a first-order view of which nodes are within direct earshot of which — without hand-correlating RX_LOG entries.

### Documentation

New page: `docs/docs/repeater-neighbors.md` (sidebar position 2.5), covering the per-repeater enable flow, the two sensors, persistence semantics, integration-level auto-cleanup, and use cases.

### Related / upcoming

Companion PRs from this series: `feature/per-device-online-sensor`, `feature/message-store`. The sidebar-panel frontend that consumes all three is tracked separately.

### DIFF
--- a/custom_components/meshcore/coordinator.py
+++ b/custom_components/meshcore/coordinator.py
@@ -49,6 +49,13 @@ from .const import (
     RATE_LIMITER_REFILL_RATE_SECONDS,
     RX_LOG_CACHE_MAX_SIZE,
     RX_LOG_CACHE_TTL_SECONDS,
+    NEIGHBOR_PUBKEY_PREFIX_LENGTH,
+    NEIGHBOR_STALE_THRESHOLD,
+    SEEN_WINDOW_SECS,
+    CONF_REPEATER_NEIGHBORS_ENABLED,
+    CONF_AUTO_CLEANUP_STALE_NEIGHBORS,
+    CONF_STALE_NEIGHBOR_DAYS,
+    DEFAULT_STALE_NEIGHBOR_DAYS,
 )
 from .meshcore_api import MeshCoreAPI
 
@@ -88,6 +95,11 @@ class MeshCoreDataUpdateCoordinator(DataUpdateCoordinator):
 
         # Storage for discovered contacts
         self._store = Store[dict[str, dict]](hass, 1, f"meshcore.{config_entry.entry_id}.discovered_contacts")
+        # Storage for neighbor data (persists SNR, seen_timestamps, etc. across restarts)
+        self._neighbor_store = Store[dict[str, dict]](
+            hass, 1, f"meshcore.{config_entry.entry_id}.neighbor_data"
+        )
+        self._neighbor_data_loaded = False
         # Get name and pubkey from config_entry.data (not options)
         self.name = config_entry.data.get(CONF_NAME)
         self.pubkey = config_entry.data.get(CONF_PUBKEY)
@@ -176,6 +188,22 @@ class MeshCoreDataUpdateCoordinator(DataUpdateCoordinator):
         # safety-net poll only fires after MSG_SAFETY_NET_INTERVAL of silence.
         self._last_msg_activity: float = 0.0
         self._initial_drain_done: bool = False
+
+        # Repeater neighbor tracking
+        # Key: repeater pubkey_prefix, Value: dict of neighbor data keyed by neighbor pubkey
+        # Each neighbor entry: {pubkey, snr, secs_ago, last_updated, resolved_name}
+        self._repeater_neighbors: Dict[str, Dict[str, dict]] = {}
+        # Track which neighbor sensor entities have been created: set of "repeater_pubkey:neighbor_pubkey"
+        self._created_neighbor_sensors: set = set()
+
+        # Auto-cleanup of stale neighbors (daily)
+        self._auto_cleanup_stale_neighbors = config_entry.data.get(
+            CONF_AUTO_CLEANUP_STALE_NEIGHBORS, False
+        )
+        self._stale_neighbor_days = config_entry.data.get(
+            CONF_STALE_NEIGHBOR_DAYS, DEFAULT_STALE_NEIGHBOR_DAYS
+        )
+        self._last_stale_neighbor_cleanup = 0.0
 
         # RX_LOG correlation cache: auto-evicts after TTL expires
         # Key: correlation hash, Value: list of RX_LOG data (multiple receptions possible)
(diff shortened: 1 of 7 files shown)


custom_components/meshcore/coordinator.py
```diff
@@ -536,13 +564,390 @@ def update_telemetry_settings(self, config_entry: ConfigEntry) -> None:
...
+                self.logger.info(
+                    "Loaded persisted neighbor data for %d repeaters (%d total neighbors)",
+                    len(stored),
+                    sum(len(n) for n in stored.values()),
+                )
+        except Exception as ex:
+            self.logger.error("Error loading persisted neighbor data: %s", ex)
+        self._neighbor_data_loaded = True
+
+    def remove_single_neighbor(self, repeater_pubkey: str, neighbor_pubkey: str) -> int:
```
awolden: This is defined but nothing calls it. PR body mentions a WebSocket API but I don't see `websocket_api.async_register_command` anywhere in the diff. Either wire it up or drop the method (and update the PR body to match).

custom_components/meshcore/sensor.py
```diff
@@ -1493,7 +1542,239 @@ def extra_state_attributes(self) -> Dict[str, Any]:
...
+            hours = secs_ago // 3600
+            mins = (secs_ago % 3600) // 60
+            last_seen = f"{hours}h {mins}m ago"
+        else:
+            days = secs_ago // 86400
+            hours = (secs_ago % 86400) // 3600
+            last_seen = f"{days}d {hours}h ago"
+
+        # Re-resolve name on each attribute read (picks up newly discovered contacts)
+        new_name = self.coordinator.resolve_neighbor_name(self._neighbor_pubkey)
```
awolden: Mutating `_attr_name` inside `extra_state_attributes` doesn't trigger a state write, so the friendly name only updates on the next unrelated state push (~2h at the default cycle). Cleaner is to subscribe to CONTACTS events in the coordinator and call `async_write_ha_state()` when a name resolves. Same pattern on the Seen sensor below.

custom_components/meshcore/translations/en.json
```diff
@@ -89,6 +89,7 @@
           "repeater_name": "Available Repeaters",
           "repeater_password": "Repeater Password",
           "telemetry_enabled": "Enable Telemetry Polling (automatically polls and discovers telemetry data)",
+          "neighbors_enabled": "Enable Neighbor Entities (creates SNR and activity sensors for each neighbor seen by the repeater)",
```
awolden, line 92: Can we add a `data_description` here (mirroring what `auto_cleanup_stale_neighbors` does below) that notes the airtime cost and firmware requirement? Project convention puts prereqs in the UI, not just docs.

awolden (commented):
Nice work on this. Defaulting it off is the right call.

Main concern is airtime visibility. Each enabled repeater adds a binary request and a response that grows with the neighbor count on every poll cycle. A user flipping this on across their repeaters has no way to see what it's costing and the impact to the broader mesh

Left a few inline comments for the other stuff. Overall shape is good, just want the cost visible before this ships widely.

mwolter805:
Thanks for the thorough review — all four points addressed in the latest push. Quick summary and then per-point detail:

**Summary of changes**

- `coordinator.py` — removed `remove_single_neighbor` (dead on this branch; lives on the sibling `feature/sidebar-panel` branch where its WebSocket wiring actually exists)
- `sensor.py` — moved the neighbor friendly-name refresh out of `extra_state_attributes` and into a `_handle_coordinator_update` override on both `MeshCoreNeighborSensor` and `MeshCoreNeighborSeenSensor`
- `translations/en.json` — added a `data_description` block with a `neighbors_enabled` entry on all four repeater flow steps (`config.add_repeater`, `config.edit_repeater`, `options.add_repeater`, `options.edit_repeater`) so the airtime cost and firmware prereq are visible at the toggle
- `docs/docs/repeater-neighbors.md` — added an **Airtime cost** section and removed the (now stale) **Removing a single neighbor** section
- PR body — dropped the "Single-neighbor removal (WebSocket API)" paragraph so the description matches the diff

Changes were also deployed on my HA instance (2026.4.3) — integration loads clean, no errors in the system log, existing sensors unaffected.

---

**1. Airtime visibility (top-level).** Added two disclosures:

- **At the toggle**, a `data_description.neighbors_enabled` helper string reading: *"Queries this repeater for its neighbor list on each poll cycle and exposes each neighbor as two sensors. Airtime cost is one direct request and response per cycle — at the default 2-hour Update Interval that's roughly one exchange every 2 hours per enabled repeater, comparable to one extra telemetry request. Requires MeshCore firmware ≥ 1.14.0."*
- **In the docs**, a new *Airtime cost* section with per-cycle byte counts (~10-byte request, ~11 bytes per neighbor in the response), the routing framing (direct/path-routed `BINARY_REQ`, not flood), hop-count scaling (`2 × (N + 1)` on-air events per cycle), cadence at the default 2-hour interval, a shared-rate-limiter note (the neighbor request consumes one token from the same `rate_limiter_tokens` bucket used by repeater status polls, login retries, and tracked-node telemetry — an empty bucket skips the cycle and retries on the next poll), and the firmware prereq.

The key framing point: the request/response is a direct path-routed binary request (`PacketType.BINARY_REQ`), not a flood advert, so the cost scales with hop count to the specific repeater rather than mesh size.

**2. `remove_single_neighbor` dead code + PR body mismatch.** Removed. You were right that there's no `websocket_api.async_register_command` on this branch — the WS handler lives on `feature/sidebar-panel` with a byte-identical copy of the coordinator method. I dropped both the method and the PR body paragraph, and removed the matching "Removing a single neighbor" section from the docs page. The sidebar-panel branch will carry the WS command and its coordinator-side counterpart into its own PR; they're out of scope here.

**3. Friendly-name refresh lag in `sensor.py`.** Moved the re-resolve out of `extra_state_attributes` and into a `_handle_coordinator_update` override on both neighbor sensor classes:

```python
@callback
def _handle_coordinator_update(self) -> None:
    """Handle coordinator updates — re-resolve friendly name before state write."""
    new_name = self.coordinator.resolve_neighbor_name(self._neighbor_pubkey)
    if new_name != self._resolved_name:
        self._resolved_name = new_name
        self._update_friendly_name()
    super()._handle_coordinator_update()
```

`CoordinatorEntity`'s base handler writes state on every coordinator tick; the override re-resolves the name first so `_attr_name` is current before the write. Name updates now propagate at the coordinator cadence (`DEFAULT_SCAN_INTERVAL`, ~30s) rather than waiting for the next repeater poll (default 2h). The side-effect-in-a-property anti-pattern in `extra_state_attributes` is gone.

I considered your alternative (coordinator-side subscription to `CONTACTS` events dispatching to sensors) and went with the sensor-side override instead: no other sensor in this repo uses cross-entity event wiring, the coordinator doesn't need to maintain a `{pubkey: sensor_refs}` map, and name re-resolution is cheap. Happy to switch to the event-subscription approach if you'd rather.

**4. Missing UI disclosure for `neighbors_enabled`.** Covered by the `data_description` additions in item 1. I followed the `auto_cleanup_stale_neighbors` pattern you called out — added the `data_description` block in all four flow steps so the helper text appears whether the user enables it during initial setup or later via Configure.

Happy to iterate on any of these — particularly the airtime copy length or the refresh-refactor approach if you prefer the event-subscription variant.

awolden (approved):
All four review points addressed. Sensor-side refresh pattern is fine — follows the repo convention and works. Thanks for the thorough response.

PR #215, opened 2026-04-23
Expose SNR and hop count on incoming direct-message events

mwolter805:
### Summary

Adds per-message signal metadata to the `meshcore_message` event payload for incoming direct messages: `hop_count` (always present) and `snr` (V3 firmware only). Both are derived from the underlying `CONTACT_MSG_RECV` packet that already carries them — this PR exposes them on the public event bus.

### Why

Companion integrations and automations that track per-contact link quality currently have to either correlate against `meshcore_delivery_update` (which only fires for messages with multi-hop RX_LOG records) or fall back to repeater-table sensors that aren't message-keyed. Surfacing `hop_count` and `snr` directly on the message event closes that gap for the single-hop / direct-DM case where no RX_LOG would be emitted.

### What changes

- `custom_components/meshcore/logbook.py` — adds `hop_count` and conditional `snr` to the `meshcore_message` payload built in `handle_contact_message` for incoming DMs.
- Adjacent comment block above the `EVENT_MESHCORE_DELIVERY_UPDATE` fire sites is updated to document the correlation-field pattern that companions key on.

23-line diff, single file.

### Stability and compatibility

Pure additive change — existing automations and companions ignore unknown fields per HA convention. The two new fields are documented as **experimental** in the companion integration API page (sibling PR) until they ship on a stable release.

### Related

Sibling branches landing alongside this one:
- `feature/structured-query-services` — three new typed services (`get_contacts`, `get_channels`, `trace`).
- `docs/companion-integration-api` — new docs page formalizing the public surface for companion integrations.

### DIFF
--- a/custom_components/meshcore/logbook.py
+++ b/custom_components/meshcore/logbook.py
@@ -246,6 +246,10 @@ async def _collect_incoming_rx_logs(
                 "message": base_event_data.get("message"),
                 "timestamp": base_event_data.get("timestamp"),
             }
+            # Correlation fields: every meshcore_delivery_update carries entity_id,
+            # sender_name, message, and timestamp — enough for downstream
+            # listeners to correlate back to a previously-received
+            # meshcore_message event without re-hashing the message text.
             hass.bus.async_fire(EVENT_MESHCORE_DELIVERY_UPDATE, update_data)
 
     except Exception as ex:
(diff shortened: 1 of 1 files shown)


awolden (approved):
verified. channel messages embed snr/rssi/path_len through `rx_log_data` correlation in `_collect_incoming_rx_logs`, `handle_contact_message` fires flat with no equivalent. adding `hop_count` and V3-only `snr` direct from `CONTACT_MSG_RECV` is the right shape.

`0xFF / negative / 0 → 0` mapping is fine, direct neighbors and "no path bytes processed" both collapse to 0 and the real distinction (added vs discovered, flood vs fixed path) lives on the contact record.

worth noting `path_len` here is a byte count over 1-byte hashes (`path_hash_mode = 0`), today they're equal so the comment is accurate. if the wire format ever shifts to 2-byte hashes worth revisiting. not a reason to hold.

lgtm

PR #216, opened 2026-04-23
Add structured query services: get_contacts, get_channels, trace

mwolter805:
### Summary

Three new services under the `meshcore.*` domain that return typed responses for companion integrations and automations to consume directly — no string-scraping `execute_command` output, no private-attribute access on the coordinator.

| Service | Purpose | Response shape |
|---|---|---|
| `meshcore.get_contacts` | Device's known contacts (added + discovered) | `{contacts: [{adv_name, public_key, pubkey_prefix, type, added_to_node, out_path_len, out_path, out_path_hash_mode}]}` |
| `meshcore.get_channels` | Configured channels | `{channels: [{channel_idx, channel_name, shared_secret_present}]}` (the shared secret itself is never returned) |
| `meshcore.trace` | Path-trace to a contact | `{trace: {hops, path, round_trip_ms, final_snr, tag}}` on success; `{trace: null, error: "<code>"}` on any failure |

All accept an optional `entry_id` for multi-device installs.

### Why

Companion integrations (such as [MeshCore-HA-UI](https://github.com/Ratty7198/MeshCore-HA-UI) and [meshcore-ha-chat](https://github.com/mwolter805/meshcore-ha-chat)) currently reach into `coordinator._channel_info` / `coordinator.get_all_contacts()` / direct SDK calls to populate UI surfaces. That's brittle (private attributes can move) and forces companions to re-implement filtering and validation that already exists in the integration. These three services move the logic into a stable, typed surface.

`trace` is a full port of `ws_trace` from the sidebar-panel branch (15s flood-discovery floor, pre-registered `PATH_RESPONSE` listener, round-trip 1-byte-hash path construction per `Mesh.cpp:41-66`, `added_to_node` gate, structured `{trace: null, error: "..."}` failure shape).

### What changes

- `custom_components/meshcore/const.py` — three `SERVICE_*` constants.
- `custom_components/meshcore/services.py` — three handlers + `_resolve_coordinator` helper + registration; ~+395 lines net.
- `custom_components/meshcore/services.yaml` — UI descriptions for the new services.
- `tests/conftest.py` — adds `binary_sensor` to the mock list so `services.py` imports under test.
- `tests/test_query_services.py` — 19 new tests covering happy paths, every error branch, and the exact byte construction for round-trip path discovery.

5 files, +1210 lines.

### Testing

- `pytest tests/` — **53/53 pass** (19 new + 34 existing) in 0.14s.

### Stability and compatibility

[...]

PR #217, opened 2026-04-23
docs: add Companion Integration API page

mwolter805:
### Summary

New documentation page (`docs/docs/companion-integration-api.md`, 158 lines) that publishes the stable public surface of `meshcore-ha` for companion integrations to build against — events, services, entity patterns, plus a deprecation policy and response-shape tables for the three structured services from the sibling PR.

### Why

Multiple companion integrations now consume `meshcore-ha`'s event bus and services:

- [MeshCore-HA-UI](https://github.com/Ratty7198/MeshCore-HA-UI) — sidebar UI consuming the event bus and `send_*` services.
- [meshcore-ha-chat](https://github.com/mwolter805/meshcore-ha-chat) — sidebar chat panel + persistent message store; uses all four events plus the new structured query services.

Today there's no published commitment about which fields and services will remain stable, which are experimental, or how breaking changes will be communicated. Companions reverse-engineer the surface from source, with no signal that a maintainer-driven change won't break them. This page formalizes the contract.

### What this page covers

- **Scope and stability** — additive vs. breaking changes, the one-release-cycle deprecation rule, what counts as "experimental".
- **Events** — `meshcore_message`, `meshcore_delivery_update`, `meshcore_message_sent`, `meshcore_connected` / `meshcore_disconnected`, plus a note on `meshcore_raw_event` being explicitly experimental.
- **Services** — full table including the three structured services from the sibling PR (`get_contacts`, `get_channels`, `trace`), with response-shape tables for each.
- **Entities** — naming patterns and how attributes vs. patterns differ in stability.
- **Example pattern** — tracking message delivery via `(entity_id, timestamp)` correlation.
- **Reference implementations** — links to the two consuming companions.
- **Deprecation and reporting** — how breaking changes get communicated; how to file regressions or request promotion of internal surfaces.

### Notes

The page intentionally does **not** document `hass.data["meshcore"][entry_id]` direct coordinator access. That's an internal surface companions currently use as a workaround; documenting it would commit upstream to maintaining it. Companions needing functionality only available via that path are pointed at the issue tracker.

### Stability of the page itself

Each documented surface gets a stability label (`stable` or `experimental`). The page itself becomes the source of truth — every change to a documented surface should ship with a corresponding update + change-log entry per the rules in the page.

### Related

Sibling branches landing alongside this one:
- `feature/dm-signal-metadata` — SNR + hop_count enrichment on incoming DM events.
- `feature/structured-query-services` — the three structured services this page documents.

### DIFF
--- a/docs/docs/companion-integration-api.md
+++ b/docs/docs/companion-integration-api.md
@@ -34,24 +34,56 @@ Fired when meshcore-ha receives a direct or channel message — the primary even
 | `channel` / `channel_idx` | string / int | channel only | Channel name + 0–255 index. |
 | `receiver_name` | string | DM only | Local device's advertised name. |
 | `rx_log_data` | array | channel, when RX_LOG correlated | Per-repeater detail (`snr`, `rssi`, `path_len`); see [Events](./events#meshcore_message). |
+| `domain` | string | always | Always `"meshcore"`. Internal; companions can ignore. |
+| `outgoing` | bool | outgoing only | Present and `true` on outgoing fires; absent on incoming. |
+| `hop_count` | int | DM only | Number of hops the message traversed. **experimental** |
+| `snr` | float | DM only, V3 firmware | Signal-to-noise ratio of the inbound DM. **experimental** |
 
-Recent releases also add `hop_count` (always) and `snr` (V3 firmware only) on direct-message payloads — **experimental** until shipped on a stable release.
+For outgoing channel messages, this event re-fires once after RX_LOG collection completes (~4 s typical) carrying the final `rx_log_data`, `repeater_count`, and `progressive: false`. Treat the re-fire as the terminal state — see the example below.
 
 ### `meshcore_delivery_update`
 
-Fired progressively as late-arriving `RX_LOG` data is correlated to an already-emitted `meshcore_message`. Companions update existing UI rather than re-firing their own event. Carries `entity_id`, `sender_name`, `message`, `timestamp` (matching the parent), plus `rx_log_data` (cumulative, not delta), `repeater_count`, `progressive` (`true` for intermediate, `false` for final), and `outgoing` (present + `true` for outgoing only). Key on `(entity_id, timestamp)` to match. See [Events](./events#meshcore_delivery_update) for optional fields.
+Fired only for *intermediate* updates while late-arriving `RX_LOG` data is being correlated to an already-emitted `meshcore_message`. The terminal update is delivered as a `meshcore_message` re-fire (see above), not a delivery_update.
+
+| Field | Type | Notes |
+|---|---|---|
+| `entity_id`, `sender_name`, `message`, `timestamp` | (parent types) | Match the parent `meshcore_message`. Use `(entity_id, timestamp)` to correlate. |
+| `rx_log_data` | array | Cumulative (not delta) per-repeater detail since the parent event. |
+| `repeater_count` | int | `len(rx_log_data)`. |
+| `progressive` | bool | Always `true`. The terminal `progressive: false` arrives on `meshcore_message`, not here. |
+| `outgoing` | bool | Present and `true` on outgoing-channel updates; absent on incoming. |
+
+Additional fields from the parent `meshcore_message` (e.g., `domain`, `message_type`, `channel_idx`) pass through unchanged.
 
 ### `meshcore_message_sent`
 
-Fired when one of the `send_*` services successfully transmits. Mirrors `meshcore_message` shape. Most companions don't need it — the existing message event is enough.
+Fired when one of the `send_*` services successfully transmits. Distinct shape from `meshcore_message` — companions tracking sent-state should subscribe here, not infer from the message event.
+
+| Field | Type | When | Notes |
+|---|---|---|---|
+| `message` | string | always | Message text. |
+| `device` | string | always | The originating config-entry id. |
+| `message_type` | `"channel"` \| `"direct"` | always | Discriminator. |
+| `receiver` | string | always | DM: contact's `adv_name`. Channel: `"channel_<idx>"`. |
+| `timestamp` | int (seconds) | always | Unix epoch — *not* ISO 8601 like `meshcore_message`. |
+| `send_id` | string (8 hex) | always | Per-send identifier; useful for correlating progressive re-fires on the message event. |
+| `contact_public_key` | string (hex) | DM only | Full public key — *not* the 12-char prefix. |
+| `ack_received` | bool | DM only | Whether the device received an ACK before the suggested timeout. |
+| `channel_idx` | int 0–255 | channel only | Channel index. |
+| `send_timestamp` | int (seconds) | channel only | Timestamp the device used in the broadcast (may differ from `timestamp` due to clock skew). |
 
 ### `meshcore_connected` / `meshcore_disconnected`
 
-Fired when the device connection comes up or goes down. Payload is the config-entry identifier — use to gate UI on connectivity.
+Fired when the device connection comes up or goes down.
+
+- `meshcore_connected`: `{"connection_type": "<usb|tcp|ble>"}`.
+- `meshcore_disconnected`: `{}` on a clean disconnect, or `{"unexpected": true}` when the SDK has given up reconnecting.
+
+Neither payload includes a config-entry identifier today. Multi-device companions can't disambiguate which entry connected/disconnected from the payload alone — read `coordinator.api.connected` after the event fires, or correlate with other state.
 
 ### Raw SDK events
 
-The integration also re-fires every `meshcore_py` SDK event as `meshcore_raw_event`. **Experimental** — schema follows the SDK, not meshcore-ha. Diagnostics only.
+The integration also re-fires every `meshcore_py` SDK event as `meshcore_raw_event` with wrapper `{event_type: str, payload: dict, timestamp: float}`. The wrapper is meshcore-ha's; only the inner `payload` follows the SDK's schema. **Experimental** — diagnostics only.
 
 ## Services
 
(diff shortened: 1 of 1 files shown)


awolden (changes requested):
deprecation policy + experimental-vs-stable distinction is exactly what companion authors need. holding for one revision pass to align with what `feature/structured-query-services` (#216) actually returns, dont want to publish a stability commitment we dont ship.

### 1. summary table, field names abbreviated inconsistently with detail tables

```
| `meshcore.get_contacts` | … | `{contacts: [{name, pubkey_prefix, type, ...}]}` |
| `meshcore.get_channels` | … | `{channels: [{idx, name, shared_secret_present}]}` |
| `meshcore.trace`        | … | `{trace: {hop_count, path, rtt_ms, ...}}` |
```

detail tables further down use the real names (`adv_name`, `channel_idx`, `channel_name`, `hops`, `round_trip_ms`). either drop the summary-table response shapes entirely or rewrite to match. preference for matching, each row is otherwise the only spec a reader sees before scrolling.

### 2. trace error codes don't match implementation

doc says: `flood_discovery_timeout`, `flood_discovery_error`, `trace_timeout`. `async_trace_service` actually returns `path_discovery_failed`, `path_discovery_rejected`, `path_discovery_timeout`, `timeout`, plus catch-alls (`send_failed`, `await_failed`, `internal_error`).

suggested replacement:
```
no_coordinator, not_connected, contact_not_found, contact_not_on_device,
contact_missing_pubkey, path_discovery_failed, path_discovery_rejected,
path_discovery_timeout, timeout, send_failed, await_failed, internal_error
```

### 3. trace success shape

doc has it right in the code block, just confirm summary table row above gets brought into sync.

### 4. DM hop_count / snr paragraph

#215 just merged so wording can be present-tense. drop "Recent releases also add" and fold into the field table directly with their own "experimental" rows. companion authors scanning the table will miss a footnote-style sentence below it.

### 5. send_message field

worth one explicit line that the service accepts **either** `node_id` (advertised name) **or** `pubkey_prefix` (≥6 hex chars). single-line description glosses over the disjunction and companions will wonder which is canonical.

none of this is structural, single revision pass.

awolden:
follow-up after a deeper pass against the codebase, few more items worth fixing in the same revision.

### 6. binary_sensor.meshcore_<device>_messages doesn't exist

`get_channel_entity_id` and `get_contact_entity_id` (`utils.py:82-94`) always include either `ch_<idx>` or the contact's pubkey before the `_messages` suffix, nothing creates a bare per-device messages sensor. the auto-entities pattern users actually rely on is the wildcard `binary_sensor.meshcore_*_messages`, which matches both per-contact and per-channel rows. drop this row or rewrite as the wildcard.

### 7. pubkey_prefix width is conflated

events column says "string (hex, 12 chars)", correct: `logbook.py:339` does `event_data.get("contact_public_key", "")[:12]`. but the entities section says "first 6 chars of the device pubkey", and the table placeholder `<pubkey>` lives in the same paragraph. contact prefix in entity IDs is `[:6]` (`binary_sensor.py:347`), in events `[:12]`. worth one explicit sentence: 12 hex chars in events, 6 in entity_ids.

### 8. (entity_id, timestamp) is not stable across devices

doc says "Stable correlation key across events." `timestamp` is `datetime.now().isoformat()` at fire time (`logbook.py:304, 360, 403`), not a packet-level timestamp. multi-device install or any cross-device automation correlating on `(entity_id, timestamp)` will not see the same value on each side. example pattern works *within a single integration instance*, worth that qualifier.

### 9. python example leaks `pending` for DMs

`pending` keyed on `meshcore_message`, only popped on the `progressive: false` branch of `meshcore_delivery_update`. delivery_update fires only from the channel message path (`_collect_incoming_rx_logs` and the channel branch of `handle_outgoing_message`). DMs never produce a delivery_update, every DM key in `pending` lives forever. either change the example to a `cachetools.TTLCache(...)` or scope the comment to channel messages.

### 10. receiver_name semantic differs by direction

`receiver_name` is set on both incoming and outgoing DMs but means different things. incoming DM (`logbook.py:301`): local device's name. outgoing DM (`logbook.py:356`): recipient's `adv_name`. doc's "Local device's advertised name" is right for incoming, wrong for outgoing. either qualify ("local device on incoming, recipient on outgoing") or just say "DM, semantic depends on direction".

### 11. hop_count / snr are incoming-DM only

PR #215 added the fields in `handle_contact_message` only, `handle_outgoing_message` direct branch doesn't set them. doc paragraph "on direct-message payloads" reads as both directions, tighten to "incoming direct messages".

### 12. deprecation policy commits to a CHANGELOG that doesn't exist

`grep -rn -i CHANGELOG` returns nothing in repo root. if we keep this policy we should ship a `CHANGELOG.md` (or commit to release-note entries on GitHub Releases) before the doc page lands, otherwise the contract is unenforceable on day one.

### retraction

didn't verify on the first pass: `meshcore_connected` and `meshcore_disconnected` do fire (`meshcore_api.py:185, 218, 269`). original review didn't claim otherwise but want it on record so this comment doesnt get read as questioning their existence.

none of #6-12 are structural, can fold into the same revision pass as the field-name and error-code reconciliation.

mwolter805:
Thanks for the careful review — and apologies, the doc shipped with more
gaps than just the ones called out. All five of your items are addressed:

1. **Summary-table response shapes** now match the detail tables
   (`adv_name`, `channel_idx`/`channel_name`, `hops`/`round_trip_ms`).
2. **Trace error-code list** now lists the 12 codes the service actually
   returns (`path_discovery_*` replacing `flood_discovery_*`, `timeout`
   replacing `trace_timeout`, plus the catch-alls).
3. **Trace success shape** was already accurate in the code block;
   summary-table row sync is covered by item 1.
4. **DM `hop_count` / `snr`** are now table rows in the `meshcore_message`
   field table with explicit "experimental" markers — no more
   footnote-style sentence below the table.
5. **`send_message` recipient line**: exactly one of `node_id` (advertised
   name) or `pubkey_prefix` (≥6 hex chars).

Your review prompted me to do a more thorough pass over every surface
on the page, and it turned up several other items the original draft
was wrong or silent about. Those are all on the same branch as well:

- **`meshcore_delivery_update`** — original wording said `progressive`
  is `true` for intermediate, `false` for final. In practice
  delivery_update only fires for *intermediate* updates (always
  `progressive: true`); the terminal update is delivered as a
  `meshcore_message` re-fire with `progressive: false`. Replaced the
  prose blob with a field table and corrected the example code below
  it — the previous example's "final pass" branch never fired and
  would have leaked entries forever.
- **`meshcore_message_sent`** — said it "mirrors `meshcore_message`
  shape," but the shape is meaningfully different: `device` instead
  of `entity_id`, `receiver` instead of `sender_name` /
  `receiver_name`, `timestamp` is int seconds (not ISO 8601),
  `contact_public_key` is the full key (not the 12-char prefix), plus
  `ack_received` / `send_id` / `send_timestamp` / `channel_idx` only
  appear here. Replaced with a field table.
- **`meshcore_connected` / `meshcore_disconnected`** — said the payload
  is the config-entry identifier; it isn't. Connected carries
  `{connection_type: "<usb|tcp|ble>"}` and disconnected carries `{}`
  or `{unexpected: true}`. Documented accurately and noted that
  multi-device companions have to correlate via other state.
- **`meshcore_raw_event`** — clarified that the wrapper
  `{event_type, payload, timestamp}` is meshcore-ha's; only the inner
  `payload` follows the SDK schema.
- **`trace` success shape** — `path` was documented as `[hex, ...]`
  but it's actually `[{hash, snr}, ...]` with the final entry omitting
  `hash` (that's the local device receiving the echo). Also marked
  `final_snr` as `float | None` — it's `None` when the path is empty
  for 0-hop direct reception.
- **`trace` error codes** — added a note that firmware-supplied error
  strings may also appear as the `error` value when the radio rejects
  the request, so the documented list isn't exhaustive. Also documented
  the additional fields some failures carry (`reason` on
  `path_discovery_failed` / `path_discovery_rejected`, `round_trip_ms`
  on `timeout`).
- **`get_contacts` / `get_channels`** — added the failure-envelope
  shapes (`{contacts: [], error: "..."}` /
  `{channels: [], error: "..."}`) so companions know to check `error`
  before consuming the data.
- **Entities** — removed `binary_sensor.meshcore_<device>_messages`
  from the patterns table (no such entity is created by the
  integration; only per-contact and per-channel indicators exist).
  Also clarified that the `<pubkey>` part of the per-contact entity_id
  is the *first 6 chars* of the contact's pubkey — not the 12-char
  `pubkey_prefix` companions know from events. Companions parsing
  entity_ids would have tripped on this.
- **`meshcore_message` field table** — added rows for `domain` (always
  present, internal) and `outgoing` (present and `true` on outgoing
  fires), and noted that outgoing channel messages re-fire this event
  after RX_LOG collection completes with the terminal `progressive: false`
  plus final `rx_log_data` and `repeater_count`.
- **`execute_command`** — noted the `command` parameter (required) and
  the optional `node_id` / `pubkey_prefix` scoping.

Net change vs the original PR commit: +68 / −16 lines. Total page is
now ~211 lines.

Verified against `services.py`, `services.yaml`, `logbook.py`,
`meshcore_api.py`, `binary_sensor.py`, and `utils.py` at v2.6.0.
Happy to walk through any of the additions, or to trim further if
anything reads as over-documentation.

awolden (approved):
re-read against source. page is in great shape, all 5 original items fixed, additional surfaces you found (delivery_update intermediate-only, message_sent shape, connected/disconnected payloads, trace `path[*]` shape, failure envelopes for `get_contacts`/`get_channels`, 6-vs-12-char pubkey distinction) all accurate. python example is materially better, original would have leaked entries forever; rewrite handles channel terminal re-fire correctly and flags DM/incoming-channel cases as needing companion-side retention.

retraction and three leftovers, none merge-blocking, posting so we can fold into a follow-up rather than another revision pass:

**retraction**: my earlier comment claimed `receiver_name` was outgoing-DM only. wrong, `handle_contact_message` (incoming DM) sets `receiver_name = DEFAULT_DEVICE_NAME` (`logbook.py:301`). both directions populate it; meaning differs (incoming = local device name, outgoing = recipient's `adv_name`). current doc row "DM only, Local device's advertised name" is right for incoming, slightly misleading for outgoing.

**leftovers**:

1. `hop_count` / `snr` are incoming-DM only. PR #215 added them in `handle_contact_message` only, `handle_outgoing_message` doesn't set them. doc table rows say "DM only" without direction qualifier. tighten to "incoming DM only".

2. python example DM retention. footnote calls out that DMs need companion-side retention, example doesn't show how. one-line `from cachetools import TTLCache; pending = TTLCache(maxsize=2048, ttl=600)` swap would make the example self-contained. optional.

3. "change log" terminology. deprecation policy refers to "the change log" without a `CHANGELOG.md` in repo. could clarify it means GitHub Releases notes. minor.

approving so this can land alongside the v2.6.0 services it documents.

PR #218, opened 2026-04-23
Add conditional logic for recipient type selection

aredon:
Makes for a cleaner user experience when using this card. Conditions hide the channel selector when "send to" is "contact" and hide the contact selector when "send to" is "channel".

Also makes this part of the docs match the "overview".

### DIFF
--- a/docs/docs/dashboard/basic-node.md
+++ b/docs/docs/dashboard/basic-node.md
@@ -38,19 +38,19 @@ sections:
               - entity: select.meshcore_recipient_type
                 name: Send To
               - type: conditional
-              conditions:
-                - entity: select.meshcore_recipient_type
-                  state: Channel
-              row:
-                entity: select.meshcore_channel
-                name: Channel
-            - type: conditional
-              conditions:
-                - entity: select.meshcore_recipient_type
-                  state: Contact
-              row:
-                entity: select.meshcore_contact
-                name: Contact
+                conditions:
+                  - entity: select.meshcore_recipient_type
+                    state: Channel
+                row:
+                  entity: select.meshcore_channel
+                  name: Channel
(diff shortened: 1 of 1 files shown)


awolden (changes requested):
conditional UX is the right idea but the YAML in this diff isnt valid. HA's `conditional` row needs `conditions:` and `row:` at the same indentation as `type:` (sibling of `type:`, not the `- type:` dash), and both new list items need to share the surrounding dash indent.

current:
```yaml
              - type: conditional
              conditions:        # wrong indent (sibling of `- type:` instead of `type:`)
                - entity: select.meshcore_recipient_type
                  state: Channel
              row:               # same problem
                entity: select.meshcore_channel
                name: Channel
            - type: conditional  # wrong dash indent (12 spaces vs 14)
              conditions:
                ...
```

fixed:

```yaml
            entities:
              - entity: select.meshcore_recipient_type
                name: Send To
              - type: conditional
                conditions:
                  - entity: select.meshcore_recipient_type
                    state: Channel
                row:
                  entity: select.meshcore_channel
                  name: Channel
              - type: conditional
                conditions:
                  - entity: select.meshcore_recipient_type
                    state: Contact
                row:
                  entity: select.meshcore_contact
                  name: Contact
              - entity: text.meshcore_message
                name: Message
```

worth pasting back into your dashboard locally to confirm it loads, `type: conditional` inside an entities card is finicky and silently renders broken if the nesting is off.

aredon:
Gotta love copy pasting yaml turning the indentation into absolute soup. Added the correction but please review again anyway.

awolden (approved):
parses cleanly, 4 entities (recipient-type select, two conditional rows, message text). lgtm

PR #226, opened 2026-04-30
Fix RX_LOG parsing for multi-byte path sizes

PacoX:
This PR fixes RX_LOG parsing for MeshCore packets using the new packed path byte format. It correctly reads the hop count, supports 1/2/3/4-byte node addresses, and restores rx_log_data population for GroupText packets.

It also keeps the parsing logic aligned with meshcore_py so raw RX log events are decoded consistently across protocol modes.

### DIFF
--- a/custom_components/meshcore/mqtt_uploader.py
+++ b/custom_components/meshcore/mqtt_uploader.py
@@ -1183,14 +1183,26 @@ class MeshCoreMqttUploader:
                 if not hash_value:
                     has_transport = route_type in (0x00, 0x03)
                     offset = 1 + (4 if has_transport else 0)
-                    path_len_value = packet_bytes[offset] if len(packet_bytes) > offset else 0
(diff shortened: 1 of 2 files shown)


awolden (approved):
cross-checked the bit layout against `meshcore_py/reader.py`, every site there does `plen >> 6` for hash mode and `plen & 0x3F` for length, matches what this PR computes.

mqtt_uploader hash-construction site (replay key for trace packets) now uses unpacked `hop_count` instead of the raw byte. for the current 1-byte-hash mode the values are identical, no behavioral break for real traffic.

`parse_rx_log_data` walks path bytes in `path_hash_size`-byte chunks (`step = path_hash_size * 2` over the hex string), correct. `0xFF` (the flood sentinel meshcore_py special-cases) would compute a 252-byte path and get rejected by the existing `len(packet_bytes) < path_end + 3` length guard, safe.

lgtm

Issue #242, opened 2026-05-17
rx_log_data often not added into event_type: meshcore_message


WJ4IoT:
Despite this message comes with hops > 0 the rx_log_data is hardly added into event_type: meshcore_message.

The settings (if relevant)
- Dockered Home assistant on NUC5i5
- Meshcore 2.6.0
- Companion Wifi connected
- Adaptive Channel Message Delivery not enabled
- No MQTT-broker in integration

RikoDEV:
This appears to be related to my https://github.com/meshcore-dev/meshcore-ha/issues/237.

I've created my fork that (should) **fix the missing `rx_log_data`** issue. Let me know if you still have missing rx_log_data.

> **Caution**: These changes were "vibecoded" — Python and Meshcore communication logic are not my main area of expertise. Test thoroughly and use at your own risk, but it fixes described issue in my HA setup.

https://github.com/RikoDEV/meshcore-ha

### Installation

1. Download this repository (or clone it).
2. Copy the `meshcore` folder to `/config/custom_components/meshcore` (create the folder if it doesn't exist).
3. Restart Home Assistant.

WJ4IoT:
Thanks a scoped region could well be the root-cause of this problem due to the extra 4 bytes transport_codes inserted after the 1ste byte in RAW (source: https://docs.meshcore.io/packet_format/).

Let's check if you were also able to add this scope region to the data

RikoDEV:
> Let's check if you were also able to add this scope region to the data

Done, but it requires adding scopes in `flood_scopes` setting first.
https://github.com/meshcore-dev/meshcore-ha/compare/main...RikoDEV:meshcore-ha:feat/expose-scope

Example rx_log_data:
```json
[
  {
    "channel_hash": "ca",
    "channel_idx": 2,
    "channel_name": "#bot",
    "flood_scope": "#waw",
    "path": "fec62d42b3",
    "path_len": 5,
    "region_scope": true,
    "route_type": 0,
    "route_typename": "TC_FLOOD",
    "rssi": -52,
    "snr": 11.5,
    "text": "PL-WAW-AntBot: @[Domino🧊]\nA24D: WX_Grzybowa_RPT🌞\nA8A0: WN Kabaty ☀️\nCAFE: PL-WAW-Ant-N-RPT",
    "timestamp": 1779303029
  }
]
```

WJ4IoT:
Great!!

[...]

Issue #237, opened 2026-05-10
[Bug] rx_log_data is empty when a message is sent using a region scope


RikoDEV:
I created a bot for my community using automation, and during testing we encountered an issue with region scope handling in this integration.

If a user sends a message with region scope, the data from `trigger.event.data.rx_log_data` is empty in the `meshcore_message` event with the `channel` message_type. When sending the same message without a region scope, I have access to path data and other information from `rx_log_data`.

Example automation to replay:

```
alias: "MeshCore Bot: Trace Route"
description: Responds to `es trace` on channel with hop/signal info
triggers:
  - trigger: event
    event_type: meshcore_message
    event_data:
      message_type: channel
      channel_idx: 2
conditions:
  - condition: template
    value_template: >-
      {{ (trigger.event.data.message | default('') | lower).startswith('es
      trace') }}
actions:
  - delay:
      seconds: 2
  - variables:
      d: "{{ trigger.event.data }}"
      rx0: >
        {{ d.rx_log_data[0] if (d.rx_log_data is defined and d.rx_log_data |
        length > 0) else {} }}
      rssi_val: "{{ rx0.rssi if rx0.rssi is defined else 'n/a' }}"
      snr_val: "{{ rx0.snr if rx0.snr is defined else 'n/a' }}"
      path_hex: "{{ rx0.path if rx0.path is defined else '' }}"
      path_len: "{{ rx0.path_len if rx0.path_len is defined else 0 }}"
      path_step: >
        {% set pl = path_len | int %} {% set hex_len = path_hex | length %} {{
        ((hex_len // pl) if pl > 0 else (6 if hex_len % 6 == 0 and hex_len > 0
        else (4 if hex_len % 4 == 0 and hex_len > 0 else 2))) }}
      path_nodes: |
        {% set step = path_step | int %}
        {% if path_hex | length >= 2 %}
          {% set ns = namespace(nodes=[]) %}
          {% for i in range(0, path_hex | length, step) %}
            {% set ns.nodes = ns.nodes + [path_hex[i:i+step]] %}
          {% endfor %}
          {{ ns.nodes | join(' → ') }}
        {% else %}
          ⚡ direct or empty
        {% endif %}
      reply_text: >
        📡 @[{{ d.sender_name }}] RSSI: {{ rssi_val }} dBm | SNR: {{ snr_val }} dB Hops: {{ path_len }} | Path: {{ path_nodes }}
  - data:
      channel_idx: 2
[...]

Issue #198, opened 2026-04-05
Multibyte hash support


jongdesteven:
Latest 2.3.5 version of meshcore-py supports multibyte hash. 
Request to implement this.

sentrymaster:
Also having issues with this. Multi-byte messages have no RX_LOG_DATA and raw SDK event does not decrypt. Following this to see when it’s updated :)

awolden:
multibyte RX_LOG support shipped in v2.7.0 through #226 and #249. closing this as implemented.

PR #246, opened 2026-05-21
fix: region-scope (TC_FLOOD) RX_LOG correlation

RikoDEV:
## Resolves
https://github.com/meshcore-dev/meshcore-ha/issues/242
https://github.com/meshcore-dev/meshcore-ha/issues/237
https://github.com/meshcore-dev/meshcore-ha/issues/198

## Summary

- **Region-scope messages (TC_FLOOD, route_type 0) were never correlating with RX_LOG data.** TC_FLOOD and TC_DIRECT (route_type 3) carry a 4-byte transport code between the header byte and the path byte. Both `parse_rx_log_data` and `parse_and_decrypt_rx_log` were reading the path byte at the wrong offset, corrupting path parsing and preventing channel hash matching for all region-scoped messages.
- **Added a fast path in both parse functions** that uses the SDK's pre-parsed `MeshcorePacketParser` fields (available when `payload_type` is present in the payload dict) — these are always correct regardless of route type, eliminating the need for manual byte-offset arithmetic for the common case.
- **Channel info is now loaded eagerly on startup** so RX_LOG decryption works immediately without waiting for the first coordinator update cycle.
- **Extended RX_LOG cache TTL** from 5 s to 30 s to reduce missed correlations.
- **Fixed AES plaintext parsing offset** — the decrypted block contains a 1-byte flags field after the 4-byte timestamp; message text now correctly starts at byte 5 instead of byte 4.

### DIFF
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -475,7 +475,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     if connected and api.mesh_core:
         try:
             _LOGGER.info("Loading channel info on startup for RX_LOG correlation...")
-            await coordinator._fetch_all_channel_info()
+            await coordinator.fetch_all_channel_info()
             _LOGGER.info(f"Startup channel info loaded: {len(coordinator._channel_info)} channels")
         except Exception as ex:
             _LOGGER.error(f"Error loading channel info on startup: {ex}")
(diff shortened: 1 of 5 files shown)


awolden (changes requested):
direction is right, AES offset matches firmware (`uncrypted[5:]` after the 4-byte timestamp and 1-byte flags), and the SDK fast path is a nice cleanup over the manual byte-offset arithmetic.

couple of asks before merge:

- `coordinator._fetch_all_channel_info()` from `__init__.py:474` is calling a private. either make it public or wrap it. also worth thinking about whether sequential `get_channel(idx)` over `_max_channels` adds noticeable startup latency on BLE/serial, since this runs before platform setup.
- TTL bump 5s -> 30s is fine, but `RX_LOG_CACHE_MAX_SIZE` is still 100. on a busy mesh that's the bound that matters, not the TTL. probably ok in practice.
- no regression test in the PR. worth adding one that round-trips a real firmware-format plaintext (`ts4 + flags1 + msg`) so the offset can't quietly slip again.

closes #237, #242, #198. lgtm once the above is addressed.

Issue #229, opened 2026-05-02
Remove old devices in HA


DuschdrBabbe:
Title: Cannot remove stale devices (repeaters / companions) after removing them from integration

Description:

After removing a repeater and a companion device from the integration, the corresponding devices still appear in Home Assistant and cannot be deleted — the delete button is simply not present.

Steps to reproduce:

	1.	Add a repeater or companion device via the integration
	2.	Remove the device from the integration
	3.	Navigate to the device in HA → Settings → Devices & Services
	4.	Notice the device still exists, with no option to delete it

Expected behavior:
Device is removed from HA (or at least a delete button is available to do so manually)

Actual behavior:
Device persists in HA UI with no delete button

awolden:
fixed by #247 and #297, both included in v2.9.0. devices that are no longer configured can now be removed from the device page.

PR #247, opened 2026-05-23
feat: allow removing orphan repeater and client devices from HA UI

jagoda:
Implements async_remove_config_entry_device so the Home Assistant UI's "Delete" button works on MeshCore repeater and tracked-client devices.

Previously, removing a tracked client (or repeater) from the integration options flow updated the config entry but left the corresponding device and its entities in the registries, with no way to remove them short of deleting the integration entirely.

The new hook:
- refuses removal of the main hub device (identifier equals the config entry id) -- the hub goes away only when the integration is removed
- refuses removal of repeater/client devices whose pubkey_prefix is still present in CONF_REPEATER_SUBSCRIPTIONS / CONF_TRACKED_CLIENTS
- allows removal of orphan devices (pubkey_prefix no longer in config) and of any unknown device carrying this domain

Home Assistant handles cascading removal of associated entities once the hook returns True.

### DIFF
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -21,6 +21,7 @@ from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.components.http import StaticPathConfig
 from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers import issue_registry as ir
+from homeassistant.helpers.device_registry import DeviceEntry
 
 
 from .const import (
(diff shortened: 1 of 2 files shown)


awolden (changes requested):
direction is right and closes #229, but two real bugs to fix before merge.

1. **pubkey length mismatch.** config stores `pubkey_prefix` truncated to 12 chars (`config_flow.py:552,685,743,763`), but `telemetry_sensor.py:524` and `device_tracker.py:183` build device identifiers with the full pubkey hex. so a repeater's telemetry/GPS device ends up as `{entry_id}_repeater_{64chars}` while the stats sensor is `{entry_id}_repeater_{12chars}`. the `pubkey_prefix in repeater_prefixes` check fails for the 64-char variant and a live, valid device becomes deletable. same hazard for clients.

   fix is something like `pubkey_prefix[:12] in repeater_prefixes`.

2. **`contact` and `unknown` node types fall through to `return True`.** `_get_node_info` returns those for nodes that aren't tracked repeaters/clients (`telemetry_sensor.py:308,314`, `device_tracker.py:139,145`), and they create real device entries. with the matcher only checking `repeater`/`client`, those devices are unconditionally deletable while the integration is still actively populating them. they'll reappear, but with possibly stale entity references in the meantime.

   either refuse removal for those types while they're producing live data, or check the coordinator's active/discovered list before allowing.

jagoda:
Thanks for the feedback @awolden. I think I've addressed your comments. Please let me know what you think.

awolden (approved):
lgtm. one thing worth a follow-up, not blocking.

steady-state coverage looks right. small edge case: when `coordinator is None` or `coordinator.data` is falsy (post-reload, after a failed refresh, during reconnect), `live_contact_prefixes` ends up empty and `contact`/`unknown` devices fall through to `return True`. narrow window in practice but the `test_missing_coordinator_orphan_contact_allowed` test enshrines the permissive behavior. worth an `or not coordinator.last_update_success` guard as a follow-up if it ever surfaces.

closes #229.

PR #248, opened 2026-05-23
execute_command: normalize plain-dict and None SDK returns

jagoda:
The req_*_sync family in the meshcore SDK returns plain dicts on success and None on timeout, rather than the Event-shaped objects most send_* / set_* commands return. The response handler only formatted the Event shape, so:
  * dict returns were silently dropped (response = None)
  * None returns surfaced as a confusing "expected dictionary, got NoneType" error in the HA UI

Pass dict returns through with bytes->hex conversion, and surface None returns as {"error": "no_response", "command": "..."} so callers always get a structured response.

### DIFF
--- a/custom_components/meshcore/services.py
+++ b/custom_components/meshcore/services.py
@@ -731,9 +731,10 @@ async def async_setup_services(hass: HomeAssistant) -> None:
                                 coordinator.async_set_updated_data(updated_data)
 
                     # Normalize the SDK return value into a JSON-safe response.
-                    # Three shapes are possible:
+                    # Possible shapes:
                     #   * Event with .payload dict — send_* / set_* commands
                     #   * Plain dict — req_*_sync (awaited response payload)
+                    #   * list / scalar / str — wrapped as {"result": <value>}
                     #   * None — req_*_sync on timeout / no response
                     if hasattr(result, "payload") and isinstance(result.payload, dict):
                         response = {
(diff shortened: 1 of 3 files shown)


awolden (changes requested):
direction is right but a couple of things to address.

- the normalizer handles Event, dict, and None, but several `req_*_sync` helpers return non-dict non-None. `req_telemetry_sync` returns a list (`payload["lpp"]`), `req_mma_sync` returns `mma_data`, `req_acl_sync` returns `acl_data` (meshcore-py `commands/binary.py:69,103,133`). those still hit the trailing `_LOGGER.info("Command result: %s", result); return` and the response is dropped. worth wrapping as `{"result": result}` or similar so they surface too.

- the new tests don't actually run. `tests/test_execute_command.py` can't be collected because `services.py` imports `from .binary_sensor import create_contact_sensor` at module top, and `conftest.py` doesn't mock `custom_components.meshcore.binary_sensor`. collection fails before any test runs. need to add `binary_sensor` (and the other sibling modules services.py imports) to `_MOCKS` in `conftest.py`. once that's in place the tests will actually exercise the change.

`{"error": "no_response", "command": ...}` is fine as a shape, just worth documenting in `events.md` since it's a new contract for callers using `return_response=True`.

jagoda:
Thanks @awolden. I believe I have addressed the first issue. However, I was unable to reproduce the test issue. `pytest tests` executes all tests in my environment. Also, I see `binary_sensor` already present in `_MOCKS` [here](https://github.com/meshcore-dev/meshcore-ha/blob/main/tests/conftest.py#L32). What else am I missing?

awolden (approved):
my bad on the test-collection complaint last round — your `conftest.py` already mocks `binary_sensor` and the suite runs clean. list/scalar/str wrap covers every current `req_*_sync` shape. lgtm.

PR #250, opened 2026-05-29
feat(services): add scope parameter to send_channel_message

RikoDEV:
## Summary

Adds an optional `scope` parameter to the `meshcore.send_channel_message` service, enabling region-scoped TC_FLOOD broadcasts.

## Changes

The service now accepts a `scope` string:

  ```yaml
  service: meshcore.send_channel_message
  data:
    channel_idx: 0
    message: "Hello Warsaw region!"
    scope: "waw"          # also accepts "#waw" — normalised automatically by the SDK
```

- Calls set_flood_scope on the device before sending, then always resets it to no scope via finally — even if the send fails.
- The scope value is included in the resulting meshcore_message_sent event, so automations can inspect which scope was used.
- Omitting scope (or passing null) sends a normal global flood — fully backwards compatible.

### Incoming scope decode flow

Received TC_FLOOD packets (`route_type == 0`) are matched against the configured allowlist. Each rx_log_data entry gains a flood_scope field with the matched name (e.g. "#waw") or null.

Matching replicates the firmware's `TransportKey::calcTransportCode`:
1. Derive a 16-byte key per allowlist entry: `SHA256("#name")[:16]`
2. Compute `HMAC-SHA256(key, [payload_type_byte] + pkt_payload)[:2]` as uint16 LE
3. Apply firmware edge-case clamping: `0 → 1, 0xFFFF → 0xFFFE`
4. Compare against the packet's transport code — first match wins

## Documentation

  - services.md — new parameter table entry, global vs. scoped usage examples, note about the set/reset lifecycle.
  - events.md — scope, send_timestamp, and send_id fields documented for meshcore_message_sent.
  - services.yaml — UI field with description and example added to the service form.
  - translations/en.json — label and description for the new flood scope allowlist option in integration settings.

## Breaking changes

None. scope is optional; all existing automations continue to work unchanged.

### DIFF
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -37,6 +37,7 @@ from .const import (
     CONF_LIMIT_DISCOVERED_CONTACTS,
     CONF_MAX_DISCOVERED_CONTACTS,
     DEFAULT_MAX_DISCOVERED_CONTACTS,
+    CONF_FLOOD_SCOPES,
     CONF_MESSAGES_INTERVAL,
     DEFAULT_UPDATE_TICK,
     REPAIR_PUBKEY_CHANGED,
@@ -48,6 +49,8 @@ from .mqtt_uploader import MeshCoreMqttUploader
 from .services import async_setup_services, async_unload_services
 from .utils import (
     create_message_correlation_key,
+    load_flood_scope_keys,
+    match_flood_scope,
     parse_and_decrypt_rx_log,
     parse_rx_log_data,
     sanitize_event_data,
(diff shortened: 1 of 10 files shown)


awolden (approved):
lgtm. couple of follow-ups, not blocking.

**concurrency caveat.** `set_flood_scope` → `send_chan_msg` → `set_flood_scope("")` is a 3-step sequence that mutates global device state, and the integration today doesn't serialize commands at the api layer. another in-flight command (coordinator telemetry, repeater status pull, another service call) can interleave between set and send and observe the scoped state or clobber it. real fix is wrapping every `mesh_core.commands.X` callsite through a single lock at the api layer. that's a broader refactor than this PR. tracking separately.

**reset hardening.** the `finally` reset isn't wrapped in its own try/except. if the reset call itself raises (timeout, disconnect mid-send) the device stays scope-set. wrapping the inner call in try/except + warning log is a one-liner — worth a follow-up.

**docs nit.** PR body says scope appears on `meshcore_message`. the code actually fires `meshcore_message_sent`, which is what `docs/docs/events.md` already documents. just worth updating the description before merge so the commit message reflects reality.

closes the gap on #242.

PR #251, opened 2026-05-30
feat: optional self-diagnostic sensors for the companion node

mwolter805:
### Summary

Adds a default-off **Self Diagnostics** option that polls the locally-attached companion node's own statistics (`get_stats_core` / `get_stats_radio` / `get_stats_packets`) on a configurable interval and exposes them as main-device sensor entities — giving the companion the same kind of diagnostic detail a managed repeater already has.

Crucially, these are **local-transport queries** to the attached radio (a 2-byte `GET_STATS` frame with no destination contact), so the poll adds **no mesh traffic** and consumes no airtime or duty-cycle — unlike the repeater status polling it mirrors.

### What it adds

- A `Self Diagnostics Enabled` toggle + `Self Diagnostics Interval` (60–3600 s, default 300 s) in the USB/BLE/TCP setup steps and in Options → Global Settings, with labels and descriptions in `translations/en.json`.
- A gated self-stats poll in the coordinator, mirroring the existing self-telemetry poll. Wrapped in try/except with debug logging; no return-value handling needed (the SDK dispatches `STATS_CORE/RADIO/PACKETS` events).
- ~14 main-device sensor entities, created **only when the option is enabled**, that subscribe to those events: uptime, TX queue length, noise floor, last RSSI/SNR, TX/RX airtime, and packet counters (received, sent, flood/direct TX, flood/direct RX, receive errors). Definitions reuse the existing repeater sensor classes/units/icons for consistency.
- Three diagnostic `problem` binary sensors decoding the radio `errors` bitmask (Packet Pool Exhausted / CAD Timeout / RX-Start Timeout). `STATS_CORE.errors` is a latching fault-flag bitmask, not a count, so it is surfaced as boolean problem flags rather than a numeric sensor (`on` = the fault has occurred since the radio last booted).
- README + docs entry describing the option, the sensors, and the fault flags.

### Design notes

- **Off by default** — existing installs gain no new entities until a user opts in (no surprise entity proliferation / recorder load).
- **Battery is not duplicated** — the companion already exposes battery voltage/percentage from the `BATTERY`/`SELF_INFO` events; `STATS_CORE.battery_mv` is ignored.
- Unit conversions match the repeater sensors (seconds → minutes for uptime/airtime). `recv_errors` is guarded for the legacy 26-byte `STATS_PACKETS` frame (reported as unknown rather than failing).
- All sensor keys carry an `entity.sensor.<key>.name` translation. Most reuse existing repeater keys; the genuinely new `tx_airtime` key got its own `name` entry (`TX Airtime`) so it renders with a friendly name rather than the bare device name. The three fault binary sensors carry `entity.binary_sensor.<key>.name` entries.

### Testing

- Deployed to a live HA instance (HA 2026.5.4, `meshcore` SDK 2.3.7) on a USB-attached companion.
- With the option off: no new entities; integration loads cleanly.
- With the option on: all ~15 sensors created and populated. Verified one `STATS_CORE/RADIO/PACKETS` event of each type carried the expected fields, and that converted uptime/airtime values matched the raw `get_stats_*` readings.
- Confirmed the poll issues only local radio queries — no mesh-packet transmission, no path/rate-limiter activity.
- Toggling the option off removes the entities from active updates on reload; toggling back on restores them.

### Notes for reviewers

- The change is self-contained and mirrors the established repeater-sensor and self-telemetry patterns.
- A follow-up could tag the new counters `EntityCategory.DIAGNOSTIC` (the repeater sensors currently don't); left as-is here for parity, open to reviewer preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,21 @@ When enabled (off by default), the integration automatically uploads repeater an
 - Requires private key export on firmware (`ENABLE_PRIVATE_KEY_EXPORT=1`)
 - Replay protection and signature verification built-in
 
+## Self Diagnostics
+
+When enabled (off by default), the integration polls the locally-attached companion node's own statistics and exposes them as sensor entities — giving the companion the same rich diagnostic tiles a managed repeater has. Enable it in the integration's **Global Settings** (or during initial setup).
+
+- **No mesh traffic.** The polls are local queries to the attached radio (`get_stats_core` / `get_stats_radio` / `get_stats_packets`) — they add no LoRa-mesh traffic and consume no airtime or duty-cycle.
+- **Off by default.** No new entities are created until you opt in, so existing installs are unaffected.
+- **Poll interval.** Configurable from 60 to 3600 seconds (default 300 s / 5 minutes).
+- **Entities created (~14 sensors).** Core: uptime, TX queue length. Radio: noise floor, last RSSI, last SNR, TX airtime, RX airtime. Packets: received, sent, flood/direct TX, flood/direct RX, receive errors. Battery is not duplicated — the companion already exposes battery voltage and percentage.
+- **Radio fault flags (3 `problem` binary sensors).** The radio's `errors` field is a bitmask of dispatcher fault events, not a count, so it is decoded into three diagnostic binary sensors with `device_class: problem`:
+  - **Packet Pool Exhausted** — the packet buffer pool ran out and a packet was dropped.
+  - **CAD Timeout** — Channel Activity Detection stayed busy too long (channel congested, or the radio may be wedged).
+  - **RX-Start Timeout** — the radio failed to (re)enter receive mode (possible radio hang).
+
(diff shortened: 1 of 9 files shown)


awolden (approved):
lgtm. couple of nits worth a follow-up, not blocking:

- `uptime` (`sensor.py:152`) is `state_class: MEASUREMENT` but it's monotonic. probably wants `TOTAL_INCREASING` for proper long-term stats.
- several of the 13 numeric stat sensors lack explicit `name` entries in `translations/en.json`, so the UI will show snake_case keys (`noise_floor`, `last_rssi`, `nb_recv`, etc) instead of readable labels.
- `data_description` for `self_diagnostics_enabled` is duplicated 4x (USB/BLE/TCP/Options).
- the 3 sequential `await` calls in the diagnostics block could be `asyncio.gather` to halve RTT on slow BLE.

PR #183, opened 2026-03-30
Add stale discovered contact cleanup with archival

mwolter805:
### Summary

Adds automatic and manual cleanup of stale discovered contacts. Over time, the discovered contacts list grows with nodes that are no longer in range or active. This feature lets users remove contacts whose `lastmod` timestamp exceeds a configurable age threshold, while preserving contacts that have been saved to the node.

Removed contacts are archived to persistent storage so they can be reviewed or cleared later via WebSocket API.

### Features

- **`cleanup_stale_discovered_contacts` service** — manually trigger cleanup with an optional `days_threshold` parameter (falls back to the Global Settings value, then to 30 days). Contacts with `added_to_node=true` are always preserved.
- **Auto-cleanup option** — disabled by default. Users can enable it in Integration Options → Global Settings via "Auto-Cleanup Stale Discovered Contacts" toggle. The "Stale Contact Threshold" setting controls the age in days (default: 30, range: 1–365). When enabled, runs once per day during the coordinator update cycle.
- **Archived contacts store** — removed contacts are persisted to `meshcore.{entry_id}.archived_contacts` storage with metadata (age, reason, timestamp). Deduplication on load ensures no duplicates accumulate.
- **Exclusion list** — contacts can be excluded from future cleanup (stored alongside the archive). This supports a future restore-from-archive workflow.
- **WebSocket API** — three commands for frontends to consume the archive:
  - `meshcore/archived_contacts` — paginated browse with search
  - `meshcore/archived_contacts_stats` — summary stats (total, date range, breakdown by reason)
  - `meshcore/clear_archived_contacts` — clear all or entries older than N days
- **Batched entity removal** — stale contacts are removed in batches of 10 with `asyncio.sleep(0)` yields between batches, preventing event bus flooding that can overwhelm WebSocket clients.
- **`lastmod`-based aging** — uses the companion device's local clock (synced by HA) rather than `last_advert` which may contain timestamps from advertising nodes with incorrect clocks.

### HA Service

**`meshcore.cleanup_stale_discovered_contacts`** — Remove discovered contacts whose `lastmod` timestamp exceeds an age threshold. Contacts saved to the node (`added_to_node=true`) are always preserved. Removed contacts are archived for later review.

| Parameter | Required | Default | Description |
|-----------|----------|---------|-------------|
| `days_threshold` | No | Global Settings value (default: 30) | Remove contacts older than this many days (range: 1–365) |
| `entry_id` | No | First available device | Config entry ID for multi-device setups |

[...]

PR #252, opened 2026-05-30
fix: repair discovered-contact entity cleanup broken by the #236 unique_id migration

mwolter805:
### Background

PR #236 scoped contact-diagnostic `binary_sensor` unique_ids by `entry_id` (from a bare `pubkey_prefix` to `"<entry_id>_contact_<pubkey_prefix>"`) and migrated existing entities to the new format. That migration runs unconditionally, so every install on #236 or later is affected — single-entry and multi-entry alike.

The migration updated the entities but not the four code paths that *delete* discovered-contact entities. Those paths still looked entities up by the old bare-`pubkey_prefix` format, which post-migration never matches. The result: each cleanup removed the contact from the in-memory discovered dict and from `.storage`, but left the `binary_sensor` orphaned in the entity registry. On the next reload HA flags every one of them "no longer being provided by the meshcore integration."

### Summary

Two fixes that must ship together — the entity-lookup repair, and a companion `discard()` fix that would otherwise turn the first fix into a regression.

### The fix

**1. Align the entity lookup with the post-#236 unique_id (coordinator + services).** The four deletion paths — stale-cleanup, FIFO eviction, the single-contact `remove_discovered_contact` service, and the `clear_discovered_contacts` clear-all branch — now compute the post-#236 unique_id and resolve it with `async_get_entity_id` for an O(1) lookup. In the eviction path this also replaces an O(N) full-registry scan that compared against the old bare-prefix format.

A new orphan-sweep step walks the registry for this entry's `"<entry_id>_contact_<hex12>"` entities and removes any whose pubkey is no longer in `get_all_contacts()`, self-healing orphans that accumulated between #236 and this fix. The 12-hex suffix match is load-bearing: without it the sweep would also match the contact-selector entity (`"<entry_id>_contact_select"`). The early-returns on an empty discovered dict / empty stale set were removed so the sweep runs on every invocation — orphans live in the registry, not the dict, so those early-returns made the sweep unreachable once the dict had already been pruned by a prior daily cleanup. The end-of-run INFO log now distinguishes dict removals from registry sweeps (previously it reported the dict-removal count as if it were the entity count).

**2. Discard tracked contacts by full pubkey, not prefix.** `binary_sensor.py` keys `coordinator.tracked_diagnostic_binary_contacts` by the full 64-char `public_key`, but the eviction, stale-cleanup, and clear-all paths called `.discard()` with the 12-char prefix — a silent no-op since the eviction/clear paths landed in #150 and the stale-cleanup path inherited the pattern in #183. It was masked the whole time because entity removal at those same sites was also failing, so a re-discovered contact still had its entity present and the membership gate never mattered. With entity removal now fixed (fix 1), the membership gate in `create_contact_sensor` would block a re-discovered contact from getting a fresh `binary_sensor`. So without fix 2, fix 1 is a strict regression for users with `auto_cleanup_stale_contacts` or `limit_discovered_contacts` enabled.

Both changes are idempotent and safe: discarding a key not in the set is a no-op, and on the next HA restart the set re-initializes empty and the "create sensors for existing contacts" loop recreates an entity for every contact in the discovered dict, healing any that were missing.

### Tests

`ast.parse` clean on both modified modules; a grep confirms all three discard sites now pass the full `public_key`. Verified on a live install: a registry carrying ~790 orphaned contact entities — accumulated over ~20 days against a discovered set held near ~35 by the configured stale-cleanup and limit settings — cleared to 0 orphans in a single `clear_discovered_contacts` call. Curated/added contacts were preserved, and there were no event-loop-blocking warnings or integration errors across the restart-and-verify window.

### Notes for the maintainer

The two commits are presented squashed here; happy to split them back into the lookup fix and the discard fix as two commits if you'd prefer to review them separately. The orphan-sweep step is the one piece that does registry I/O on every cleanup invocation rather than only when the dict changed — called out in case you'd rather gate it behind a one-time migration flag instead of running it each pass.

### DIFF
--- a/custom_components/meshcore/coordinator.py
+++ b/custom_components/meshcore/coordinator.py
@@ -309,14 +309,18 @@ class MeshCoreDataUpdateCoordinator(DataUpdateCoordinator):
         for public_key in keys_to_evict:
             pubkey_prefix = public_key[:12]
             del self._discovered_contacts[public_key]
-            self.tracked_diagnostic_binary_contacts.discard(pubkey_prefix)
+            self.tracked_diagnostic_binary_contacts.discard(public_key)
 
-            for entity in list(entity_registry.entities.values()):
-                if entity.platform == DOMAIN and entity.domain == "binary_sensor":
-                    if entity.unique_id == pubkey_prefix:
-                        _LOGGER.info(f"Evicting binary sensor entity: {entity.entity_id}")
-                        entity_registry.async_remove(entity.entity_id)
-                        break
+            # Post-PR-#236 contact unique_ids are scoped by entry_id; the
+            # migration at __init__.py:_migrate_unique_ids_scope_contact_diagnostics
+            # guarantees every existing entity uses this format.
+            unique_id = f"{self.config_entry.entry_id}_contact_{pubkey_prefix}"
+            entity_id = entity_registry.async_get_entity_id(
+                "binary_sensor", DOMAIN, unique_id
+            )
+            if entity_id:
+                _LOGGER.info(f"Evicting binary sensor entity: {entity_id}")
+                entity_registry.async_remove(entity_id)
 
         _LOGGER.info(f"Evicted {evict_count} oldest discovered contacts (limit: {max_contacts})")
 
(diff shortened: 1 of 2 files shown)


awolden (approved):
lgtm. one small thing worth catching in scope since the rest of the PR fixes the same set-key bug elsewhere: `async_remove_discovered_contact_service` at `services.py:1031` doesn't call `tracked_diagnostic_binary_contacts.discard(full_pubkey)`. pre-existing on main, but the prefix→full-pubkey discard switch was the whole point here. happy to land as-is and follow up.

PR #259, opened 2026-06-04
fix: use utc timestamps for mqtt

### LINKED ISSUE #258: fix: use UTC-aware timestamps in mqtt_uploader.py
`mqtt_uploader.py` uses `datetime.now().isoformat()` in three places, which produces naive local-time timestamps with no timezone indicator (e.g. `2026-06-04T23:46:42.123456`). On any Home Assistant host running in a non-UTC timezone, these timestamps are offset from UTC but carry no suffix to indicate that — so receiving brokers have no way to interpret them correctly.

This causes the following error on MeshCore-compatible brokers (e.g. CoreScope / LetsMesh(?) ):

```
Naive observer clock — timing is being clamped
This observer is emitting zone-less local-time timestamps and its clock is
currently Xh ahead of UTC. Per-packet rxTime for this observer is being
collapsed to ingest time, which muddies propagation-delay analytics.
```

Note: `logbook.py` is unaffected — it already correctly uses `dt_util.utcnow()`.

## Fix

Import `timezone` from the standard library and pass it to all three `datetime.now()` calls in `mqtt_uploader.py`:

```python
# Before
from datetime import datetime
...
datetime.now().isoformat()   # lines 822, 1019, 1148
```

```python
# After
from datetime import datetime, timezone
...
datetime.now(timezone.utc).isoformat()   # lines 822, 1019, 1148
```

This produces offset-aware ISO 8601 timestamps (e.g. `2026-06-05T17:46:42.123456+00:00`) that brokers can unambiguously interpret as UTC.

## Files changed

- `custom_components/meshcore/mqtt_uploader.py` — 4 lines (1 import, 3 call sites)

## Testing

Reproduced on a Home Assistant host in `Europe/Vilnius` (UTC+3). After this fix, broker clamp events stop and the error banner clears within 24 hours.

Athemis:
## Summary

- Emit UTC-aware MQTT timestamps for status, raw event, and normalized packet payloads
- Add regression tests for MQTT timestamp formatting
- Update documentation

## Why

CoreScope reports and clamps MQTT observer timestamps when they are emitted as zone-less local-time values. Using UTC offset-aware ISO 8601 timestamps prevents consumers from interpreting local time as UTC and applying false clock-skew corrections.

## Testing

- `pytest tests/test_mqtt_timestamps.py -v`

Fixes #258

### DIFF
--- a/custom_components/meshcore/mqtt_uploader.py
+++ b/custom_components/meshcore/mqtt_uploader.py
@@ -50,11 +50,6 @@ def _as_int(value: str | int | None, default: int) -> int:
         return default
 
 
-def _utc_timestamp() -> str:
-    """Return an offset-aware UTC ISO-8601 timestamp."""
-    return datetime.now(timezone.utc).isoformat()
-
-
 @dataclass
 class BrokerConfig:
     """Configuration for one MQTT broker."""
@@ -824,7 +819,7 @@ class MeshCoreMqttUploader:
         """Build status payload compatible with other uploaders."""
         payload: dict[str, Any] = {
             "status": state,
-            "timestamp": _utc_timestamp(),
+            "timestamp": datetime.now(timezone.utc).isoformat(),
             "origin": self.node_name,
             "origin_id": self.public_key or "DEVICE",
             "source": "meshcore-ha",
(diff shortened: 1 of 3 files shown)


yellowcooln:
I will be loading this up in my test env tomorrow @awolden. But from the looks of this, I see no issues with a merge.

awolden:
@yellowcooln I assume your test env found no issues?

awolden (approved):
lgtm. one heads up: the same `now` also feeds the packet `time`/`date` display fields, so those flip from local to UTC wall-clock too. consistent with the fix, just worth a line in mqtt.md so nobody's surprised by the date rolling over at a different hour.

yellowcooln:
> @yellowcooln I assume your test env found no issues?

Yeah sorry for the delay, I didn't see any issues when testing.

PR #261, opened 2026-06-08
feat: collapse contact-discovery flags into one tri-state mode

mwolter805:
### Summary

This reshapes the PR per your review. It replaces the two overlapping contact-discovery flags — the existing `disable_contact_discovery` and the `large_mesh_mode` boolean this PR originally proposed — with a single **`contact_discovery_mode`** select offering three values:

- **Entity per contact** (`full`) — today's default: one `binary_sensor` per discovered contact.
- **Data only** (`data_only`) — discovered contacts are tracked as data (visible in the discovered-contact dropdown, the aggregate summary sensor, and the `get_discovered_contact` lookup service) but get no per-contact entity. This is the behaviour the `large_mesh_mode` boolean added.
- **Disabled** (`off`) — discovered-contact processing is suppressed and the existing discovered set is cleared. This is what `disable_contact_discovery=true` did, now also reconciling already-stored discovered contacts.

Added/curated contacts keep their entities in every mode. Existing installs migrate automatically via a config-entry migration (no reload, no entity-registry churn). With the mode left at its migrated default, behaviour is unchanged from today.

This answers your 2026-06-13 note that the two booleans ask the same question — "how much discovered-contact machinery do I want" — with different mechanics, and that a second overlapping flag is a confusing config surface.

### Problem

The integration creates one `binary_sensor` per discovered contact. On dense meshes this means hundreds of entities for transient peers heard in passing. On my instance, 100 of ~105 discovered-contact entities sat permanently in `discovered` — never promoted, never charted, never used in an automation — and the surfaces users actually work through (the discovered-contact dropdown, the add/remove/clear services, the chat panel) all read coordinator data rather than these entities. That is the situation Home Assistant's [entity-disabled-by-default rule](https://developers.home-assistant.io/docs/core/integration-quality-scale/rules/entity-disabled-by-default/) warns about: noisy, low-utility entities shouldn't make every user pay their tracking cost by default.

The larger cost is churn. Measured on my instance, the registry's `deleted_entities` array held **766 deleted meshcore contact entities — 94.6% of all 810 deleted entities in the entire registry** (every other integration combined: 44). That debris re-accumulated within two days of being manually cleared, and 453 of the 766 carried numeric collision suffixes (`_contact_2`, `_3`, …) — direct evidence of contacts cycling through create/delete repeatedly. The entity registry is a single multi-megabyte JSON store HA rewrites in full on a short debounce whenever entities are added or removed, so this cycle drives repeated whole-file rewrites (flash wear on the SD/eMMC typical of HA hardware) plus recorder state history for entities that live a few days.

Separately, the integration already had `disable_contact_discovery` to suppress discovered-contact processing entirely. So before this PR there were two flags addressing the same axis with different mechanics. This reshape unifies them.

### Approach

A single `contact_discovery_mode` setting (stored values `full` / `data_only` / `off`, default `full`), read through one accessor `get_contact_discovery_mode(config_entry)` that every consumer calls. The fundamental gate is unchanged from the original: `create_contact_sensor` suppresses a per-contact `binary_sensor` for contacts not in the coordinator's added set — now when the mode is `data_only` or `off` (added contacts always get an entity; only `full` creates one for a discovered contact).

Existing entries migrate in `async_migrate_entry` (config-entry `VERSION` 2 → 3), which runs before setup, rewrites `config_entry.data` only, and never touches the entity registry or schedules a reload:

- `disable_contact_discovery=true` → `off` (wins the tie-break even if `large_mesh_mode` was also set — no discovery means nothing to be data-only)
- else `large_mesh_mode=true` → `data_only`
- else → `full`

### Reconciliation of existing contacts on mode switch

Changing the mode reloads the entry, and a single reconciliation pass runs on every setup, in the coordinator's post-reload path, to enforce the chosen mode on the *existing* discovered population — so a switch takes effect immediately rather than only applying to newly-heard contacts. This restores the immediate cleanup you removed as racy, but race-free: it runs after the reload against the final coordinator/registry state, not from the options flow.

- **Disabled** (`off`) — clears the existing discovered set (persisted, so it stays empty across reloads) and gates incoming adverts so they don't repopulate it; the dropdown and summary sensor go empty. The radio keeps hearing adverts — HA just stops tracking them. It is entirely HA-side: the companion is already in manual mode, so discovered contacts only ever lived in HA, and there is no radio mutation.
- **Data only** (`data_only`) — removes the per-contact entities of existing discovered contacts but keeps their data (the dropdown, summary sensor, and `get_discovered_contact` still work).
- **Entity per contact** (`full`) — recreates per-contact entities for the existing discovered set on the next setup, and for new contacts as they are heard.

The pass is idempotent (a no-op in a steady mode) and reuses the demote sweep's unique-id allowlist, so it never touches added contacts, tracked-client/subscription entities, or repeater-neighbor sensors. This removes the earlier need to run `clear_discovered_contacts` by hand after switching.

### What's included

- **`const`** — `contact_discovery_mode` key, the three mode values, default, and the `get_contact_discovery_mode` accessor. The two old `CONF_*` flags are removed once all consumers are re-keyed; the migration reads the literal old keys so it doesn't depend on the retired constants.
- **`__init__` migration** — the `v2→v3` branch above; v1 entries chain through to v3 in one pass.
- **`binary_sensor`** — `create_contact_sensor` suppresses discovered contacts in `data_only` and `off`; both call paths (the contacts-update event path and the platform-setup path) route through the one gate, so a reload in any mode is consistent.
- **`coordinator` / `services`** — the eviction, stale-cleanup, remove-discovered, and clear-discovered paths run unconditionally; in `data_only`/`off` the registry lookup simply returns nothing, and the same pass cleans up any entities stranded by a prior mode. The demote path (removing an added contact) still removes that contact's acquired entities in `data_only`/`off`, with subscription-managed telemetry/GPS entities excluded.
- **`config_flow`** — `VERSION = 3`; one `contact_discovery_mode` SelectSelector replaces both booleans on every install path (USB/BLE/TCP) and in the options step. The racy toggle-on migration is deleted (the entry migration replaces it).
- **`sensor`** — the aggregate "Discovered Contacts" summary sensor (diagnostic, disabled-by-default) is unchanged and created in all modes, so users can gauge the discovered count before switching.
- **`services`** — `get_discovered_contact` gains a min-length guard on `pubkey_prefix` (enforced both at the `vol.Schema` and inline) so an empty prefix returns `not_found` instead of matching an arbitrary contact.
- **`README` + docs site + `en.json`** — retitled to the unified tier model (Entity per contact / Data only / Disabled), the inspection surfaces, and the one trade-off.

### What it does NOT change

- **Default behaviour** (mode `full`): byte-for-byte today's — entity per discovered contact, all cleanup paths active.
- **Added/curated contacts**: get entities in every mode; promotion creates the entity.
- **Channels, repeater/neighbor telemetry, companion sensors, message store, RX_LOG correlation, logbook, MQTT uploader**: none are per-discovered-contact; all unchanged.
- **The discovered-contact dropdown**: already data-driven from coordinator data, so it lists discovered contacts in `full` and `data_only`.
- **`meshcore_py`**: no SDK change and no `manifest.json` requirement bump — the mode gates are HA-side only.

### Review fixes (from your 2026-06-11 notes)

1. **Toggle-on migration raced the options reload** → removed entirely; replaced by the `async_migrate_entry` config-data migration, which runs before setup and never schedules a reload or touches the registry. (The entity cleanup it used to do is now handled race-free by the setup-time reconciliation pass above.)
2. **`if large_mesh` gates blocked entity removal exactly when leftovers existed** → the cleanup gates are gone; removal runs unconditionally and is a harmless no-op when there's nothing to remove.
3. **"Run `clear_discovered_contacts` after enabling" guidance** → no longer needed; mode switches reconcile existing contacts automatically (see above).
4. **Empty `pubkey_prefix` matched an arbitrary contact** (`startswith("")` is always true) → min-length guard at the schema and inline.

### Trade-off (the one visible loss)

In `data_only` and `off` there is no per-discovered-contact connectivity sensor, so no per-contact charting or automation for un-added contacts. The aggregate summary sensor and the `get_discovered_contact` service cover inspection; promoting a contact to "added" restores a full entity.

### Testing

Unit coverage drives the real code paths: the `create_contact_sensor` gate across all three modes (including a test that AST-extracts and runs the *real* function against production source, with a negative control), the `async_migrate_entry` mappings (disable→off, both→off, large→data_only, neither→full, old keys dropped, v1→v3 chained), the unconditional cleanup paths, the demote-removal decision and its telemetry/GPS sweep with subscription exclusion, the summary sensor, and the lookup service including the empty-prefix guard. Config-flow tests confirm each install path defaults to `full` and the options step round-trips the select. Full suite green on Python 3.13; `py_compile` and `en.json` / `services.yaml` parse clean. All 197 tests pass.

The change has also run on my own instance for a day of normal operation: the migrated entry stayed on `data_only` with the old keys gone, the discovered/added census stayed stable across several restarts, and there was no migration re-fire, reload-race, or new error attributable to the change.

### Notes / open questions for review

- **Labels.** Used **Entity per contact** / **Data only** / **Disabled** per your suggestion that "full" is too vague as a user-facing label; stored values stay `full` / `data_only` / `off`.
- **Storage location.** `contact_discovery_mode` is in `config_entry.data` alongside the existing contact-settings cluster, per your call; the broader `data` → `options` migration remains a separate refactor.
- **Accessor.** Added `get_contact_discovery_mode(config_entry)` as the single read point, replacing the scattered `.data.get(...)` reads.
- **Migration.** Proper `async_migrate_entry` (VERSION 2→3), before setup, no reload, no registry access — per your "migration approach is right" note; the off-wins tie-break is as you confirmed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,16 @@ A companion Lovelace card is available at [meshcore-card](https://github.com/jpe
 
 For detailed configuration instructions, see the [documentation](https://meshcore-dev.github.io/meshcore-ha/).
 
+## Contact Discovery Mode
+
+A single **Contact Discovery Mode** setting controls how much per-discovered-contact machinery the integration creates, with three choices:
+
+- **Entity per contact** (default) — every discovered contact gets its own diagnostic binary sensor, as before.
+- **Data only** — discovered (un-added) contacts are tracked as data only, with no per-contact entity, while contacts you add to your node keep their entities as usual. On dense meshes this avoids hundreds of low-utility entities and the entity-registry churn they drive.
+- **Disabled** — no discovered-contact processing at all (no contact sensors, no persistence, empty selectors) — for when you only track specific repeaters or clients.
+
+Set it at install time or later in **Configure → Global Settings**. In **Data only** mode the discovered-contact dropdown, messaging, services, and the chat panel all keep working, and the data-only contacts stay inspectable via an aggregate summary sensor and the `meshcore.get_discovered_contact` service; the only trade-off is no individual connectivity sensor / charting / automation for un-added contacts. See [Contact Management → Contact Discovery Mode](https://meshcore-dev.github.io/meshcore-ha/contacts#contact-discovery-mode).
+
 ## MQTT Upload (Addon/Container Env)
 
 Configuration can be done in the Home Assistant Web UI:
--- a/custom_components/meshcore/__init__.py
+++ b/custom_components/meshcore/__init__.py
@@ -42,6 +42,11 @@ from .const import (
     CONF_MESSAGES_INTERVAL,
     DEFAULT_UPDATE_TICK,
     REPAIR_PUBKEY_CHANGED,
+    CONF_CONTACT_DISCOVERY_MODE,
+    MODE_FULL,
+    MODE_DATA_ONLY,
+    MODE_OFF,
+    get_contact_discovery_mode,
 )
 from .coordinator import MeshCoreDataUpdateCoordinator
 from .meshcore_api import MeshCoreAPI
@@ -79,7 +84,7 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry) ->
     _LOGGER.debug("Migrating configuration from version %s", config_entry.version)
     
     # Don't allow downgrading from future versions
-    if config_entry.version > 2:
+    if config_entry.version > 3:
         _LOGGER.error("Cannot downgrade from version %s", config_entry.version)
         return False
     
@@ -108,6 +113,26 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry) ->
         
         _LOGGER.info("Migrated configuration from version %s to version 2", config_entry.version)
     
+    # Migrate from version 2 to version 3: collapse the two overlapping
+    # discovery booleans into the single contact_discovery_mode tri-state.
+    # Standalone `if` (not `elif`): a v1 entry's v1->v2 block above sets
+    # version=2, so it falls into this block and chains v1->v3 in one pass.
+    if config_entry.version == 2:
+        new_data = dict(config_entry.data)
+        # Read the legacy keys as string literals so the migration is
+        # independent of the const definitions.
+        disable = new_data.pop("disable_contact_discovery", False)
+        large_mesh = new_data.pop("large_mesh_mode", False)
+        if disable:
+            mode = MODE_OFF  # "off" wins the tie-break: no discovery, nothing to be data-only
+        elif large_mesh:
+            mode = MODE_DATA_ONLY
+        else:
+            mode = MODE_FULL
+        new_data[CONF_CONTACT_DISCOVERY_MODE] = mode
+        hass.config_entries.async_update_entry(config_entry, data=new_data, version=3)
+        _LOGGER.info("Migrated contact-discovery settings to %s (version 3)", mode)
+
     _LOGGER.debug("Migration to configuration version %s successful", config_entry.version)
     return True
 
@@ -518,7 +543,20 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     # Set up all platforms for this device
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
-    
+
+    # Bring the EXISTING discovered-contact population into line with the
+    # configured contact discovery mode now that platforms are up (entities,
+    # the binary_sensor add-callback, and the tracking sets all exist). Runs
+    # once per setup with the correct post-reload coordinator, so it covers
+    # start, reload, and the reload a mode switch triggers -- without the
+    # pre-reload race. full is a near no-op (the setup create-pass already
+    # built entities); data_only/off remove the discovered per-contact
+    # entities, and off also clears the discovered set.
+    try:
+        await coordinator.async_reconcile_discovered_for_mode()
+    except Exception as ex:
+        _LOGGER.error("Error reconciling discovered contacts to mode: %s", ex)
+
     # Register static paths for icons
     should_cache = False
     icons_path = Path(__file__).parent / "www" / "icons"
@@ -663,6 +701,13 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         async def handle_new_contact(event):
             """Handle NEW_CONTACT events for discovered but not-yet-added contacts."""
             if not event or not event.payload:
+                return
+
+            # Discovery disabled (mode off): drop incoming adverts so they do
+            # not repopulate the discovered set (mirrors the off early-return in
+            # binary_sensor.handle_contacts_update). HA-side only -- the radio
+            # still hears and forwards the advert; HA just stops keeping it.
+            if get_contact_discovery_mode(entry) == MODE_OFF:
                 return
 
             contact = event.payload
--- a/custom_components/meshcore/binary_sensor.py
+++ b/custom_components/meshcore/binary_sensor.py
@@ -20,7 +20,6 @@ from homeassistant.helpers.update_coordinator import (
 )
 
 from .const import (
-    CONF_DISABLE_CONTACT_DISCOVERY,
     CONF_SELF_DIAGNOSTICS_ENABLED,
     DOMAIN,
     ENTITY_DOMAIN_BINARY_SENSOR,
@@ -31,6 +30,9 @@ from .const import (
     SELF_DIAG_ERR_CAD_TIMEOUT,
     SELF_DIAG_ERR_POOL_FULL,
     SELF_DIAG_ERR_RX_TIMEOUT,
+    MODE_DATA_ONLY,
+    MODE_OFF,
+    get_contact_discovery_mode,
     NodeType,
 )
 from .utils import (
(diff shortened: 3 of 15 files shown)


awolden (changes requested):
looks like a good addition, couple of things to fix first:

- drop the toggle-on migration in `config_flow.py:936-955`. it races the options reload (`async_update_entry` schedules a reload, the later `hass.data` lookup can miss or grab the fresh coordinator and delete added contacts' entities)
- drop the `if large_mesh` gates in eviction, stale cleanup, and clear. they block entity removal exactly when leftovers exist, and the lookup they skip is a no-op for data-only contacts
- instead just document "run `meshcore.clear_discovered_contacts` after enabling" in the data_description
- `get_discovered_contact` with empty `pubkey_prefix` matches an arbitrary contact (`services.py:1395`), add a min-length check

this makes it all simpler and still supports the use cases

awolden:
stepping back on the design: we already have `disable_contact_discovery`, and this adds a second overlapping toggle for basically the same question (how much discovered-contact machinery do I want). two booleans with different mechanics is going to confuse.

would rather collapse to one tri-state setting:

- **Full** — entities for everything (today's default)
- **Data only** — discovered tracked as data, no entities, dropdown still works (this PR's large mesh mode)
- **Off** — no discovery (today's `disable_contact_discovery`)

that subsumes both flags with no overlap. would mean reshaping this into a mode select and migrating the old bool → enum on load.

bigger ask than the migration/gates fixes, so want your read before you sink more time in. happy to talk through it.

mwolter805:
Yeah, I'm on board with collapsing to one tri-state — you're right that two booleans with different mechanics is a confusing surface, and `disable_contact_discovery` + this are the same axis. Happy to reshape this into a mode select.

How I'd map it (sanity-check me):

A single `contact_discovery_mode` select replacing both booleans:

- **Full** — entity per discovered contact (today's default: `disable_contact_discovery=false`, `large_mesh_mode=false`)
- **Data only** — discovered tracked as data, no per-contact entities, dropdown/services still work (this PR's mode)
- **Off** — no discovery at all (today's `disable_contact_discovery=true`)

Migration: I'd do the bool→enum as a proper entry migration (version bump, runs before setup) rather than an options-flow `async_update_entry` — so unlike the toggle-on migration you flagged, there's no reload to race and no entity-registry touch. Mapping:

- `disable_contact_discovery=true` → `off` (wins if both were ever set — with discovery off there are no discovered contacts to be data-only)
- else `large_mesh_mode=true` → `data_only`
- else → `full`

…then both old keys drop out of `config_entry.data`.

On your four review points — all folding into the reshape:

- **Toggle-on entity-removal migration**: gone. Only the config bool→enum rewrite above remains, and that doesn't schedule a reload or touch the registry. For switching into Data only I'll document "run `meshcore.clear_discovered_contacts` after enabling" in the option's `data_description`, like you said.
- **`if large_mesh` gates** in eviction/stale-cleanup/clear: dropped — cleanup just runs; the lookups are harmless no-ops when there are no discovered entities.
- **`get_discovered_contact` empty prefix**: adding the min-length guard.
- The Data-only behaviors (the `create_contact_sensor` gate, the summary sensor, the lookup service, the demote-cleanup) just re-key from the boolean to `mode == data_only` — same behavior, only the condition source changes.

A few things I'd like your call on before I rework:

1. **Names** — OK with `contact_discovery_mode` as the key and Full / Data only / Off as labels? Or different wording (e.g. "Entity per contact" / "Data only" / "Disabled")?
2. **Storage** — both old booleans live in `config_entry.data`; I'd keep the enum there too for now (matches the rest of the contact-settings cluster) and leave the data→options move as a separate refactor, same as I flagged above. Good, or do you want it in `config_entry.options` as part of this?
3. **Summary sensor + lookup service** — keep them (they're the inspection path for Data only), or split either into its own PR?

If that shape looks right I'll rebuild the PR around it. Thanks for the careful review.

awolden:
migration approach is right — proper entry migration before setup, no reload, no registry touch. the off-wins tie-break is the correct call.

1. names: `contact_discovery_mode` key, labels **Entity per contact / Data only / Disabled**. "full" was too vague.
2. storage: keep the enum in `config_entry.data` with the rest of the contact-settings cluster, leave the data→options move as its own refactor. no need to blaze a new trail with one key.
3. summary sensor + lookup service: keep them in this PR — they're the inspection path for data only, without them that mode ships with no way to see discovered contacts.

shape looks right, go for it.

mwolter805:
@awolden — I've reshaped this per your review; the branch is force-updated to a single squashed commit.

The two overlapping flags (`disable_contact_discovery` and the `large_mesh_mode` this PR proposed) are now one `contact_discovery_mode` select:

- **Entity per contact** (`full`) — today's default, unchanged.
- **Data only** (`data_only`) — the large-mesh behaviour: discovered contacts tracked as data, no per-contact entity.
- **Disabled** (`off`) — what `disable_contact_discovery=true` did.

All reads go through one `get_contact_discovery_mode` accessor. Existing installs migrate via `async_migrate_entry` (VERSION 2→3) — before setup, `config_entry.data` only, no reload and no registry access — with `disable` winning the tie-break.

Your four review points are folded in rather than patched on top:

1. The racy toggle-on migration is gone, replaced by the entry migration (no reload).
2. The cleanup gates now run unconditionally (a no-op when there's nothing to remove).
3. Switching modes now reconciles existing discovered contacts automatically — the immediate cleanup you flagged as racy, restored race-free in the coordinator's post-reload path (not the options flow) — so the "run `clear_discovered_contacts` after switching" step is no longer needed.
4. `get_discovered_contact` rejects an empty/short `pubkey_prefix` at both the schema and inline.

Labels are **Entity per contact / Data only / Disabled** per your note that "full" reads as vague (stored values stay `full`/`data_only`/`off`); storage stays in `config_entry.data`, with the `data`→`options` move left as a separate refactor. The PR description above is rewritten to match. I've run this on my own instance for a day across several restarts with no regression — happy to keep iterating on naming or shape.

awolden (approved):
solid rework, this is the shape I wanted and the migration + reconcile logic both hold up under review. pushed the one blocker fix myself (the migration tests were `async def` with no `@pytest.mark.asyncio`, so all 9 were erroring instead of running; suite is 197 green now).

two non-blocking follow-ups for later:
- reconcile is only tested against a hand-written mirror, not `async_reconcile_discovered_for_mode` itself. your `test_create_contact_sensor_real.py` docstring notes the off-mode bug that shipped past a buggy mirror, same gap here.
- `_remove_discovered_contact_entities` calls `er.async_entries_for_config_entry` once per discovered contact (O(N×E) per boot on the dense meshes data_only targets). worth hoisting the entry list out of the loop.

nice work.

PR #263, opened 2026-06-09
fix: recognize '*' as the global wildcard in the flood-scope allowlist parser

mwolter805:
## Summary

The Flood Scope Allowlist parser — `load_flood_scope_keys()` in `custom_components/meshcore/utils.py`, added in #250 — mishandles a raw `*` entry. Instead of recognizing `*` as the global wildcard and skipping it, it derives and stores a junk transport key for a region `#*` that cannot exist. It's a one-line ordering bug: the parser's *own* wildcard guard runs *after* the name has already been `#`-prefixed, so `*` reaches the guard as `#*` and slips through. This is inbound-parser-only — no SDK change, no send-path change.

## Reproduction

```python
>>> from custom_components.meshcore.utils import load_flood_scope_keys
>>> list(load_flood_scope_keys("*, pl-mz").keys())
['#*', '#pl-mz']                                  # '#*' should not be here
>>> load_flood_scope_keys("*, pl-mz")["#*"].hex()
'156a1317d89806742661b1ba3ed4c6c0'                # junk key for a region that can't exist
```

A user who lists `*` in **Devices → MeshCore → Configure → Global Settings → Flood Scope Allowlist** (e.g. `*, pl-mz, pl-waw`) gets a spurious `#*` key in the inbound matcher keyset.

## Root cause

`load_flood_scope_keys()` already intends to skip the wildcard — it guards `name not in ("*", "#")` — but the guard runs *after* `_normalize_flood_scope_name()` prepends `#`:

```python
for entry in scopes_str.split(","):
    name = _normalize_flood_scope_name(entry)   # "*"  ->  "#*"
    if name and name not in ("*", "#"):          # "#*" is not in ("*", "#")  ->  passes
        result[name] = hashlib.sha256(name.encode()).digest()[:16]
```

A raw `*` therefore arrives as `#*`, passes the guard, and a junk key `SHA256("#*")[:16]` is added. A literal `#` *is* correctly skipped (normalization leaves a leading `#` untouched), which shows the guard ordering is the only defect.

## Why `#*` is wrong, not intended

- **The existing guard documents the intent** to exclude `*`; only its position defeats it.
- **`*` is the reserved wildcard / "null" region — not a hashable region.** `RegionMap::is_name_char()` rejects `*`, so a region literally named `*` cannot be created; `RegionMap::findByName()` special-cases `*` to the wildcard; and the region-filtering blog post explicitly calls `*` the "null region." So `SHA256("#*")` names no reachable region.
- **No node transmits a `#*`-scoped flood.** `set_flood_scope("*")` resets to the disabled (unscoped) state, and the inbound matcher only runs on transport-flood packets (`route_type == 0`). The junk key matches no real traffic; its only residual effect is a ~1-in-65536-per-packet transport-code collision that could mislabel an unrelated scoped flood — which the fix removes.

## Why users put `*` in the allowlist

`*` is the wildcard region every repeater carries — the natural way to write "all regions / global" for anyone coming from the repeater CLI or the mobile app. Downstream tools that build a scope picker from this allowlist want to present `*` as an explicit "all regions" choice, which the junk-key behavior undermines.

## The fix

Guard the raw, stripped token *before* normalization:

```python
# Tokens that denote the global/unscoped wildcard rather than a named region.
FLOOD_SCOPE_WILDCARD_TOKENS: Final = ("*", "0", "None", "#", "")


def load_flood_scope_keys(scopes_str: str) -> dict[str, bytes]:
    result: dict[str, bytes] = {}
    if not scopes_str:
        return result
    for entry in scopes_str.split(","):
        raw = entry.strip()
        if raw in FLOOD_SCOPE_WILDCARD_TOKENS:   # skip wildcard / disable tokens
            continue
        name = _normalize_flood_scope_name(raw)
        result[name] = hashlib.sha256(name.encode()).digest()[:16]
    return result
```

(`Final` is added to the existing `from typing import ...` line.) The token set mirrors the SDK's `set_flood_scope()` disable set (`"*"`, `"0"`, `"None"`, `""`) — capital-N `"None"` only, since region names are lowercase and `none` is a legitimate region — plus the parser-specific `"#"`. Named regions are unchanged: `_normalize_flood_scope_name` still applies and `#`-prefix idempotence is preserved.

## Compatibility

The only behavior change is that a `*` allowlist entry no longer contributes a key to the inbound matcher keyset. Named-region matching is byte-identical (see tests), and there is no effect on the send path or on any firmware/repeater behavior. A user who somehow relied on the `#*` key — impossible, since it matches no real traffic — would see no change.

## No SDK change required

This is confined to the inbound allowlist parser. The SDK already maps `*`/`0`/`""`/`None` to the 16-byte zero key (scope reset) in `set_flood_scope()`'s disable branch — confirmed against `meshcore` 2.3.7, the `manifest.json` floor:

| `scope` argument | emitted 16-byte key |
|---|---|
| `"*"` / `"0"` / `""` / `None` | `00000000000000000000000000000000` |
| `"pl-mz"` ≡ `"#pl-mz"` | `338787d625e26039aa7b14389f4f0c1a` |
| `"pl-waw"` | `004e244d7f21a728393c84793e599080` |

So no SDK change or version bump is needed.

## Tests

Added to the parser's unit coverage:

- `load_flood_scope_keys("*, pl-mz, pl-waw")` → exactly `{"#pl-mz", "#pl-waw"}` — no `#*`, no `*`.
- Each of `"*"`, `" * "`, `"0"`, `"None"`, `"#"`, `""`, `"   "` alone → empty keyset.
- Lowercase `"none"` → hashed as a normal region (`SHA256("#none")[:16]`), guarding the capital-N-only parity.
- `"#pl-mz"` and `"pl-mz"` → identical key (regression: named handling unchanged).
- Named-region keys byte-identical to the pre-change function on a fixed input (proves only wildcard handling changed).

## Optional doc note

A one-line clarification in `docs/docs/services.md`: *"in the Flood Scope Allowlist, `*` is recognized as the wildcard rather than hashed as a literal region name."*

---

Refs #250 (which introduced the allowlist and this parser).

### DIFF
--- a/custom_components/meshcore/utils.py
+++ b/custom_components/meshcore/utils.py
@@ -6,7 +6,7 @@ import hashlib
 import hmac
 import logging
 import re
-from typing import Any
+from typing import Any, Final
 
 from Crypto.Cipher import AES
 from homeassistant.util import slugify
@@ -482,19 +482,38 @@ def _normalize_flood_scope_name(name: str) -> str:
     return name
 
 
+# Tokens that denote the global/unscoped wildcard rather than a named
+# region. An unscoped flood already reaches every repeater through the
+# '*' wildcard region they all carry, so the wildcard has no transport
+# key and is never hashed into the inbound keyset. Mirrors the SDK's
+# set_flood_scope disable set ('*', '0', 'None', '') — capital-N "None"
+# only; lowercase "none" is a valid (lowercase-only) region name — plus
+# the parser-specific '#'.
+FLOOD_SCOPE_WILDCARD_TOKENS: Final = ("*", "0", "None", "#", "")
+
+
 def load_flood_scope_keys(scopes_str: str) -> dict[str, bytes]:
     """Parse a comma-separated flood_scopes string into a name→16-byte-key dict.
 
     Key derivation mirrors the firmware's auto-key for #hashtag regions:
     SHA256(scope_name_bytes)[:16].  Names are normalized to have a '#' prefix.
+    The '*' wildcard (and its empty/'0'/'#' equivalents) is the global scope,
+    not a named region — it carries no key and is skipped.
     """
     result: dict[str, bytes] = {}
     if not scopes_str:
         return result
     for entry in scopes_str.split(","):
-        name = _normalize_flood_scope_name(entry)
-        if name and name not in ("*", "#"):
-            result[name] = hashlib.sha256(name.encode()).digest()[:16]
+        raw = entry.strip()
+        # Guard the RAW token BEFORE normalization. _normalize_flood_scope_name
+        # prepends '#', turning a bare '*' into '#*' — which the previous
+        # post-normalize `name not in ("*", "#")` guard failed to catch, so a
+        # junk SHA256('#*') key leaked into the matcher. Guarding the raw token
+        # skips every wildcard form cleanly.
+        if raw in FLOOD_SCOPE_WILDCARD_TOKENS:
+            continue
+        name = _normalize_flood_scope_name(raw)
+        result[name] = hashlib.sha256(name.encode()).digest()[:16]
     return result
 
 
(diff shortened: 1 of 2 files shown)


awolden (approved):
lgtm

PR #266, opened 2026-06-17
fix: handle string-shaped public_key in _get_node_info

mwolter805:
### Summary

`TelemetrySensorManager._get_node_info` and `DeviceTrackerManager._get_node_info` both extract a contact's public key with `contact.get("public_key", {}).get("hex", "")`, assuming `public_key` is a `{"hex": ...}` dict. Coordinator contacts carry `public_key` as a plain hex string. When a `TELEMETRY_RESPONSE` / GPS event arrives for a node that is **not** a tracked repeater, **not** a tracked client, and **not** the root node, the lookup falls through to the contacts loop and raises `AttributeError: 'str' object has no attribute 'get'` on the first string-pubkey contact — killing the event handler before any sensor is created.

Net effect: telemetry / GPS sensors are never created for non-subscribed contacts (e.g. an added contact that answers Req Telemetry gets nothing), and both handlers log one ERROR per event.

### Fix

Route both copies through a new shared helper `utils.extract_pubkey_hex(contact)`:

```python
def extract_pubkey_hex(contact: dict[str, Any]) -> str:
    pk = contact.get("public_key", "")
    if isinstance(pk, dict):
        return pk.get("hex", "")
    return pk or ""
```

It returns the hex string for the current string shape and still handles the legacy `{"hex": ...}` dict shape, so the two previously-duplicated call sites can't silently drift again.

### Testing

- New `tests/test_extract_pubkey_hex.py`: covers string, legacy-dict, and missing/empty (`""`, `None`, `{}`, dict-without-`hex`) shapes; asserts a string-pubkey contact survives the `.startswith()` used in the lookup loop; and guards that both `_get_node_info` copies route through the helper.
- Full test suite green.
- Manually verified on a live instance: a Req Telemetry to a non-subscribed contact now creates its telemetry sensor (channel-1 voltage) instead of raising `AttributeError` and creating nothing.

### DIFF
--- a/custom_components/meshcore/device_tracker.py
+++ b/custom_components/meshcore/device_tracker.py
@@ -20,7 +20,8 @@ from .utils import (
     format_entity_id, 
     build_device_name, 
     get_device_model, 
-    build_device_id
+    build_device_id,
+    extract_pubkey_hex,
 )
 from . import MeshCoreDataUpdateCoordinator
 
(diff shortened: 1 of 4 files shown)


awolden (approved):
lgtm

PR #267, opened 2026-06-17
Push node_status state on event and seed initial value

mwolter805:
## Summary

After an integration restart, the header connection indicator (the `node_status` sensor) can show "Disconnected" while the API is actually connected, until the device next emits an event. Two causes in the `node_status` branch of `MeshCoreSensor.async_added_to_hass`:

1. The `update_status` callback set `_native_value` but never called `async_write_ha_state()`, so the change was not pushed — state only updated incidentally on the next coordinator poll. Every sibling callback in this method (`node_count`, `tx_power`, …) already calls it.
2. No initial value was seeded at setup, so after a restart where the API connected before any event arrived, the sensor sat at `None`.

## Changes

- `update_status` now calls `async_write_ha_state()` so the change is pushed immediately.
- The initial value is seeded from `coordinator.api.connected` in `async_added_to_hass`.

## Testing

Added `tests/test_node_status_state_push.py`, which exercises the real `node_status` branch (asserts the callback pushes state and that the initial value is seeded from the connection state). Full test suite passes (`pytest tests/`).

Also verified on a live Home Assistant instance: after deploying the change and restarting, the `node_status` sensor was set to `online` at integration setup time (its `last_changed` matched the sensor-platform setup timestamp to the millisecond), rather than sitting at `unknown` until a later event.

## Severity

Low/Medium — cosmetic but misleading; self-corrects on the next event.

### DIFF
--- a/custom_components/meshcore/sensor.py
+++ b/custom_components/meshcore/sensor.py
@@ -1025,14 +1025,15 @@ class MeshCoreSensor(CoordinatorEntity, SensorEntity):
 
         if key == "node_status":
             def update_status(event: Event):
-                if self.coordinator.api.connected:
-                    self._native_value = "online"
-                else:
-                    self._native_value = "offline"
+                self._native_value = "online" if self.coordinator.api.connected else "offline"
+                self.async_write_ha_state()
             meshcore.dispatcher.subscribe(
                 None,
                 update_status,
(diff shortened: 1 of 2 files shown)


awolden (approved):
lgtm

PR #268, opened 2026-06-17
fix: quiet benign no_event_received errors from get_msg loops

mwolter805:
## Problem

During a config-entry re-add or a radio-link reconnect, two `get_msg()` loops in `coordinator.py` can receive an `EventType.ERROR` result whose payload is `{'reason': 'no_event_received'}` — a startup/reconnect race where the flush or poll fired before the radio link produced its first event. It has no functional impact (the next cycle recovers), but both loops logged every `EventType.ERROR` result at ERROR, so the benign race surfaced as ERROR-tier log noise:

- `async_flush_messages` → `Error flushing messages: {'reason': 'no_event_received'}`
- the `_async_update_data` safety-net poll → `Error retrieving messages: {'reason': 'no_event_received'}`

## Fix

Route both sites through a shared `_log_get_msg_error(action, payload)` helper that logs the `no_event_received` reason at DEBUG and keeps ERROR — with the original message text — for every other failure, including non-dict payloads.

## Tests

`tests/test_get_msg_error_log_level.py` AST-extracts the real `_log_get_msg_error` helper and asserts `no_event_received` → DEBUG for both actions, that other reasons and non-dict payloads stay ERROR with the original text, plus a guard that both call sites route through the helper.

Verified live on Home Assistant 2026.6.2: with the radio link interrupted, the safety-net poll's `no_event_received` now logs at DEBUG (`Skipped retrieving messages, radio link still coming up`) with zero ERROR lines.

### DIFF
--- a/custom_components/meshcore/coordinator.py
+++ b/custom_components/meshcore/coordinator.py
@@ -69,6 +69,23 @@ _LOGGER = logging.getLogger(__name__)
 MSG_SAFETY_NET_INTERVAL: int = 60
 
 
+def _log_get_msg_error(action: str, payload: Any) -> None:
+    """Log a ``get_msg()`` ERROR result at the appropriate level.
+
+    The ``no_event_received`` reason is a benign startup race: a flush or poll
+    fired before the radio link produced its first event, and the next cycle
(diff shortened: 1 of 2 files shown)


awolden (approved):
lgtm

Issue #265, opened 2026-06-13
HA integration doesnt work, error: Failed to connect


Worthbiggles:
# MeshCore Home Assistant Integration BLE Connection Analysis

### Home Assistant

* Installation Method: Home Assistant OS
* Core Version: 2026.6.2
* Supervisor Version: 2026.06.1
* OS Version: 17.3
* Frontend Version: 20260527.5

### MeshCore

* Integration Version: 2.7.0 (HACS)
* Firmware Version: 1.16.0 (June 2026)
* Hardware: Heltec V4 (ESP32)

### Connection Method

* Bluetooth Low Energy (BLE)
* Device Name: MeshCore-Homebase
* Device Address: F8:5B:1B:A6:7E:25

---

## Initial Symptoms

The MeshCore integration could not be added through Home Assistant.

Initially the Bluetooth subsystem in Home Assistant reported:

> requested auto mode but is operating in passive mode. The scanner is in a bad state and needs to be power cycled.

After a full host reboot, Bluetooth returned to a healthy state and the warning disappeared.

At that point:

* BLE advertisements were visible
* The MeshCore device became visible in Bluetooth advertisements
* The device was marked as connectable

---

## Bluetooth Discovery Verification

Home Assistant successfully discovers the device:

```json
{
  "name":"F8:5B:1B:A6:7E:25",
  "address":"F8:5B:1B:A6:7E:25",
  "rssi":-83,
  "service_uuids":[
    "6e400001-b5a3-f393-e0a9-e50e24dcca9e"
  ],
  "connectable":true
}
```

The advertised service UUID is:

```text
6e400001-b5a3-f393-e0a9-e50e24dcca9e
```

which appears to be the Nordic UART Service (NUS).

This confirms that:

* BLE advertising is working
* Home Assistant receives advertisements correctly
* The device is connectable

---

## Bluetooth Connection Verification

The Home Assistant Bluetooth stack successfully establishes a BLE connection.

Relevant log excerpt:

```text
INFO [custom_components.meshcore.meshcore_api]
Using BLE connection with address F8:5B:1B:A6:7E:25

DEBUG [bleak.backends.bluezdbus.client]
Connecting to device @ F8:5B:1B:A6:7E:25

DEBUG [habluetooth.wrappers]
Connected via hci0
```

This demonstrates that:

* Device discovery succeeds
* Device selection succeeds
* BLE connection establishment succeeds
* BlueZ and Bleak successfully connect to the device

---

[...]

PR #294, opened 2026-06-24
BLE: support passkey-protected companion radios (PIN + BlueZ pairing agent)

drewmccal:
## What

Adds support for BLE companion radios that require a passkey/PIN to pair (e.g. Elecrow ThinkNode-M1). Previously these failed at connect time with `org.bluez.Error.AuthenticationFailed`.

## Why

`bleak`'s `client.pair()` needs a BlueZ pairing agent to supply the passkey. HAOS's default agent is no-input/just-works, so a radio that demands a PIN can never complete bonding from the integration — the connection just fails.

## Changes

- **`CONF_BLE_PIN`** optional field in the BLE config + reconfigure flows (blank = no PIN, normalized to `None`).
- **`ble_pairing_agent.py`** (new): registers an `org.bluez.Agent1` via `dbus_fast` (already present transitively via `bleak`/`habluetooth`, imported lazily) that answers `RequestPasskey`/`RequestPinCode`/`RequestConfirmation`/etc. with the configured PIN. Capability `KeyboardDisplay`. Registered before `create_ble`, unregistered in `finally`.
- **`meshcore_api.py`**: accepts `ble_pin`, registers/unregisters the agent around the BLE connect.
- Translations for the new field.
- Tests: PIN normalization, `create_ble` forwarding, graceful fallback when the agent can't register.

If agent registration fails, connection still proceeds (works for already-bonded / just-works radios), so this is non-regressive for existing setups.


awolden:
nice work, this solves a real gap (#265/#83). but it belongs a layer down in meshcore-py, not here.

the SDK already owns this and is just missing the last piece: `BLEConnection` takes `pin=` and calls `client.pair()` (`ble_cx.py:113-116`), and `create_ble(pin=...)` exposes it. on BlueZ, `pair()` raises a passkey request the default agent can't answer, so the SDK's `pin=` silently does nothing today. your `ble_pairing_agent.py` is exactly the missing piece that makes it work, and it's pure BLE-transport plumbing (raw system DBus, no HA APIs). every meshcore-py consumer hits the same wall: the `meshcore-cli` that's in our own manifest, bots, and the SDK's own example all call `create_ble(pin=...)` and silently fail to pair without it.

since it's not urgent, rather than carry a stopgap here and migrate later, let's do it right. filed meshcore-dev/meshcore_py#92 as the target:

- contribute the agent to meshcore-py, registered/unregistered around the `pair()` call in `ble_cx.py`, guarded by `if pin is not None` with the lazy `dbus_fast` import you already have. a test that doesn't just mock `pair()` would be good (the SDK's current test mocks it, which is why this gap went unnoticed).
- then this PR reduces to the HA-side bits, which are right: collect `CONF_BLE_PIN` in the BLE config flow and pass it through to `create_ble(pin=...)`, and update `installation.md`. drop the DBus module here.

happy to review the meshcore-py PR. the config-flow work stays.

PR #296, opened 2026-06-24
Repeater login & polling resilience: rejection vs timeout, auto path-reset, bounded mesh calls

### LINKED ISSUE #202: Login timing out with larger hop numbers despite repeater responding and providing path.
Our mesh is pretty large and covers two cities. I have a repeater (room server) I would like to monitor that is 6 hops away and roughly ~45km. When attempting to add monitoring to one such repeater the expected anon_request floods out over the mesh. The repeater responds and provides path info but not before the integration times out, and, funnily enough, the path to that contact updates for future attempts despite saying timed out. 

You can see this from the below. Five repeats is a direct path, ~20 repeats is a flood. 

Image

Even with the direct path though the integration is still timing out what seems like right before the response is received. On my personal companion when I log into this node it waits a very very long time before timing out over 6 hops (and then lets me in). These images are from different attempts but you get the idea. 

Image

drewmccal:
Closes #202

## What

Three related robustness fixes for repeater login and polling, all stemming from the same lesson: a stale stored path (or a wedged link) shouldn't look like a generic failure.

### 1. Distinguish login rejection from timeout (config flow)
`send_login_sync` used a tight (~5 s) window and only waited for `LOGIN_SUCCESS`, ignoring `LOGIN_FAILED` — so a wrong password and an unreachable repeater both surfaced as the same vague error. Now `_attempt_login` waits for *both* `LOGIN_SUCCESS` and `LOGIN_FAILED` on a wider window and reports `success` / `rejected` / `timeout` distinctly (`login_failed` vs `login_timeout` error strings).

### 2. Reset stale path → flood and retry on first timeout
A freshly-added or moved repeater has a stale direct path, so the first request silently times out and the Online sensor sits at "unknown" (gray card) until `MAX_FAILURES_BEFORE_PATH_RESET` backoff-spaced failures elapse. `_login_to_repeater` (config flow) and `_call_with_path_recovery` (coordinator) now reset the path to flood and retry **once on the first timeout** — the same thing the iOS app's manual "clear path" did. Applied to the repeater status request, recovery login, and telemetry request. Only fires when a direct path exists (`out_path_len > -1`), so it can't loop while flooding, and it honors `disable_path_reset`.

### 3. Bound every mesh call so a hang can't wedge polling
No mesh call in a repeater poll had a timeout. A wedged BLE/serial link could leave one `await` hanging forever; the poll task then never completed and the "still running, skipping" guard skipped that repeater on every tick until restart (observed with a large `fetch_all_neighbours`). Every mesh call in the poll path is now wrapped in `asyncio.wait_for` (`MESH_COMMAND_TIMEOUT` / `NEIGHBOR_FETCH_TIMEOUT`), turning a hang into an ordinary recoverable failure.

## Tests
`tests/test_path_recovery.py` covers the path-reset-on-timeout logic and the hang-treated-as-no-response behavior.
Tested with a Thinknode M1 and Raspberry Pi 4 running HA.


awolden (changes requested):
the bounded-call timeouts are good, keep those. drop the coordinator first-timeout path reset (`_call_with_path_recovery` on login/status/telemetry):

- stale paths already recover via `MAX_FAILURES_BEFORE_PATH_RESET` (3 failures); this just moves the trigger to the 1st timeout
- `reset_path` forces a flood (region-wide airtime); the 3-strike threshold is a deliberate throttle, we want fewer floods not more
- flip-flop: a successful flood-retry resets the failure counter so backoff never engages, so a marginal node re-floods nearly every poll
- reset + retry sends also bypass the rate limiter

zooming out: a lot of the integration's design is about being a good citizen on the mesh (rate limiter, backoff, the 3-strike threshold). flooding on the first timeout cuts directly against that, it makes us one of the noisier nodes in the region instead of one of the quieter ones. that's the opposite of the direction we want to push.

keep the existing 3-strike reset. rejection-vs-timeout in the add-repeater flow is fine (one-time).

the test is a standalone copy of the method and has drifted from the real one (`sleep(0)` vs `sleep(0.5)`); drive the real coordinator method instead.

PR #307, opened 2026-07-10
fix: make WebSocket path configurable for EMQX compatibility

bplein:
## Summary
  - Fixes WebSocket connection failures with EMQX broker, which serves MQTT over WebSocket at `/mqtt` instead of `/`
  - Adds a new `ws_path` broker config option (default: `/` for backwards compatibility)
  - Existing configs default to `/` on next save, maintaining compatibility with most broker deployments.
  - Allows users of different brokers to customize the MQTT path if needed

  ## Context
  The hardcoded `ws_set_options(path="/", ...)` caused the Home Assistant integration to fail when connecting to EMQX, which expects WebSocket connections at the `/mqtt` path. This is the default WebSocket
  endpoint for EMQX deployments.

  ## Changes
  - `mqtt_uploader.py`: Added `ws_path` field to `BrokerConfig` with default `/`
  - `mqtt_uploader.py`: Pass `ws_path` from broker settings when creating `BrokerConfig`
  - `mqtt_uploader.py`: Use `broker.ws_path` instead of hardcoded `/` in `ws_set_options`
  - `config_flow.py`: Added `ws_path` field to broker config UI form
  - `config_flow.py`: Persist `ws_path` value when saving broker configuration

  ## Testing
  - Verified against EMQX 6.2.0 with WebSocket at `/mqtt`
  - Backward compatible — existing configs without `ws_path` default to `/`
  - Users can override to any path via the HA integration config UI
  - **Note:** Not yet tested with non-EMQX brokers — please test with your broker and report any issues
  
  ## Admission of Guilt
  - I used Claude to develop this fix, but I do understand the relatively minor changes made to the code.
  - I have not tested it against other popular brokers.

### DIFF
--- a/custom_components/meshcore/config_flow.py
+++ b/custom_components/meshcore/config_flow.py
@@ -1099,7 +1099,7 @@ class OptionsFlowHandler(config_entries.OptionsFlow):
                 "owner_email": user_input.get("owner_email", ""),
                 "topic_status": user_input.get("topic_status", DEFAULT_MQTT_TOPIC_STATUS),
                 "topic_events": user_input.get("topic_events", DEFAULT_MQTT_TOPIC_EVENTS),
-                 "ws_path": user_input.get("ws_path", "/mqtt"),
+                "ws_path": user_input.get("ws_path", "/mqtt"),
                 "iata": user_input.get("iata", legacy_global_iata),
                 "token_ttl_seconds": user_input.get("token_ttl_seconds", legacy_global_ttl),
                 "payload_mode": user_input.get("payload_mode", "packet"),
--- a/custom_components/meshcore/config_flow.py
+++ b/custom_components/meshcore/config_flow.py
@@ -1099,7 +1099,7 @@ class OptionsFlowHandler(config_entries.OptionsFlow):
                 "owner_email": user_input.get("owner_email", ""),
                 "topic_status": user_input.get("topic_status", DEFAULT_MQTT_TOPIC_STATUS),
                 "topic_events": user_input.get("topic_events", DEFAULT_MQTT_TOPIC_EVENTS),
-                "ws_path": user_input.get("ws_path", "/mqtt"),
+                "ws_path": user_input.get("ws_path", "/"),
                 "iata": user_input.get("iata", legacy_global_iata),
                 "token_ttl_seconds": user_input.get("token_ttl_seconds", legacy_global_ttl),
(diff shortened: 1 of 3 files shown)


custom_components/meshcore/config_flow.py
```diff
@@ -1099,6 +1099,7 @@ async def async_step_mqtt_broker(self, user_input=None):
                 "owner_email": user_input.get("owner_email", ""),
                 "topic_status": user_input.get("topic_status", DEFAULT_MQTT_TOPIC_STATUS),
                 "topic_events": user_input.get("topic_events", DEFAULT_MQTT_TOPIC_EVENTS),
+                 "ws_path": user_input.get("ws_path", "/mqtt"),
```
andyshinn: Minor nit: indentation aligned on 4 space.

```suggestion
                "ws_path": user_input.get("ws_path", "/mqtt"),
```

custom_components/meshcore/mqtt_uploader.py
```diff
@@ -74,6 +74,7 @@ class BrokerConfig:
     client_id_prefix: str
     topic_status: str
     topic_packets: str
+    ws_path: str = "/mqtt"
```
andyshinn: The existing default was `/` and this changes the new default to `/mqtt`. It might make sense to change the default to `/` unless we know that `/mqtt` is what most users are already using.

bplein: That does make sense. I’ll make the change today.

bplein: There were additional places where this incorrect default was set.  Fixed and tested with https://github.com/meshcore-dev/meshcore-ha/pull/307/changes/4f79a85287abdf889cc63d23b3f05a8e4ebd3520

bplein:
Converted to draft, unexpected issue I need to resolve on my end (may not be related to the PR itself)

bplein:
Errors were configuration issues on EMQX.

awolden (changes requested):
Looks good, one thing before merge: add `ws_path` to the `mqtt_broker` step in `translations/en.json` (data + data_description), otherwise the form renders the raw key. While you're in there: `.strip() or "/"` instead of `or "/").strip()` so whitespace input falls back too, and the indentation on that `).strip()` line is off. README lines could use a quick reword and have trailing whitespace.

PR #323, opened 2026-08-09
Track advert paths and expose them on contact sensors

jpettitt:
When the companion pushes an `ADVERTISEMENT` event, the coordinator fetches the route that advert took (`GET_ADVERT_PATH` — a local device query, no RF traffic) and exposes it on the contact binary sensor as `adv_path` / `adv_path_len` / `adv_path_time`.

**Why:** `out_path` only reflects a learned return route, which stays flood-routed until a two-way exchange with the contact. The firmware records the traversed path for every advert it hears, so this makes "how was this contact last heard" visible to dashboards — e.g. meshcore-card's routing-path display, which today can only show Direct/Flood for advert-only contacts.

**Notes**
- Fetching latches off after 3 consecutive failures so firmware without `GET_ADVERT_PATH` isn't queried on every advert (retries after HA restart).
- Paths populate as adverts arrive; no persistence across restarts (the neighbor-store pattern could be added later if wanted).
- Docs: contact attribute list updated in contacts.md.
- Tests: 7 integration-tier tests driving the real coordinator and sensor with only the MeshCore API mocked.

### DIFF
--- a/custom_components/meshcore/binary_sensor.py
+++ b/custom_components/meshcore/binary_sensor.py
@@ -728,7 +728,11 @@ class MeshCoreContactDiagnosticBinarySensor(CoordinatorEntity, BinarySensorEntit
         # Add all contact properties as attributes
         for key, value in self._contact_data.items():
             attributes[key] = value
-            
+
+        # Route the last advert took (adv_path/adv_path_len/adv_path_time),
+        # tracked by the coordinator from ADVERTISEMENT pushes.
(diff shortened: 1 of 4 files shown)


awolden (changes requested):
advert-path lookups need to be serialized per companion. the response has no contact key, so two concurrent lookups can both consume the same response. I reproduced both contacts getting the same path.

please add a test with overlapping lookups for different contacts, and bound the path cache so contact churn can't grow it indefinitely.

PR #349, opened 2026-09-19
fix: do not block HA startup with RX_LOG collection task

### LINKED ISSUE #339: RX_LOG collection task delays HA startup wrap-up (tracked task instead of background task)
### Problem

In `custom_components/meshcore/logbook.py`, the adaptive-mode background RX_LOG collector is scheduled as a **tracked** task:

```python
if adaptive and hash_key is not None:
    hass.async_create_task(
        _collect_incoming_rx_logs(hass, coordinator, hash_key, event_data)
    )
```

`hass.async_create_task()` puts the task in the set that `hass.async_block_till_done()` waits on, which Home Assistant calls during the startup "wrapping up" phase before firing `EVENT_HOMEASSISTANT_STARTED`. `_collect_incoming_rx_logs()` sleeps through `_INCOMING_BG_PASS_INTERVALS` (0.5s + 1.0s), so any channel message received while HA is still starting holds up the wrap-up phase — and a steady stream of them keeps holding it up.

It showed up on my instance in the bootstrap timeout list:

```
2026-09-08 17:36:39.109 WARNING (MainThread) [homeassistant.bootstrap] Setup timed out for bootstrap waiting on {<Task pending name='Task-6140' coro=<_collect_incoming_rx_logs() running at /config/custom_components/meshcore/logbook.py:249> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[set.remove()]>, ...} - moving forward
```

This is a progressive-delivery side task whose result nothing awaits, so it has no business gating startup.

### Suggested fix

```python
if adaptive and hash_key is not None:
    hass.async_create_background_task(
        _collect_incoming_rx_logs(hass, coordinator, hash_key, event_data),
        "meshcore_rx_logs",
        eager_start=True,
    )
```

Background tasks are excluded from `async_block_till_done()`, so the collector runs exactly as it does now without participating in startup.

### Environment

- MeshCore v2.9.0
- Home Assistant Core 2026.9.1, HAOS 18.2, Python 3.14.6

HarshRajSinghania:
## Summary

Schedule adaptive incoming RX_LOG collection with `hass.async_create_background_task()` instead of `hass.async_create_task()`.

## Motivation

Fixes #339.

`async_create_task()` is tracked by `hass.async_block_till_done()`. `_collect_incoming_rx_logs()` sleeps through `_INCOMING_BG_PASS_INTERVALS` (0.5s + 1.0s), so a channel message received during Home Assistant startup wrap-up holds that phase — and a stream of them can keep holding it up. This showed up in bootstrap timeout logs on the reporter's instance.

The collector is a progressive-delivery side task; nothing awaits its result, so it should not gate startup.

## Implementation

In `custom_components/meshcore/logbook.py`, replace:

```python
hass.async_create_task(_collect_incoming_rx_logs(...))
```

with:

```python
hass.async_create_background_task(
    _collect_incoming_rx_logs(...),
    name="meshcore_rx_logs",
)
```

Background tasks are excluded from `async_block_till_done()`. The named-task pattern matches the existing ACK-wait usage in `services.py`.

`eager_start` is omitted to stay consistent with that existing call site.

## Testing

- `python3 -m py_compile custom_components/meshcore/logbook.py` — passed
- Full Home Assistant bootstrap timeout cannot be reproduced in this environment (requires HAOS + live mesh traffic during startup). The change is the API the issue requested and matches an already-tested pattern in this repo.

## Notes

No other `async_create_task` call sites were changed. Coordinator neighbor-save, MQTT auth refresh, and setup tasks are different lifetimes and were left as-is.


awolden (changes requested):
ya moving this to a background task makes sense. please also tie it to the config entry so it gets cancelled on unload, and add a test covering that.

